PR #4, opened 2018-04-10
doc/examples: remove 'ninja' as explicit dep

grmat:
as it's already required by meson.

Also use the default unit file path (/usr/lib/systemd/user/) for packaged versions. /etc/systemd/user is intended for manually installed services.

arch-meson is just a Meson wrapper with correct arch packaging prefixes set.


mdiluz:
Thanks,

Happy to merge if @ysblokje gives the OK as well.

grmat:
If you don't want to use `arch-meson` for general demonstrative reasons, @ysblokje could still use it for his AUR packages.

Issue #3, opened 2018-04-10
Error while compiling


aly199:
Got this error after running ./bootstrap.sh

+ meson --prefix=/usr build -Dwith-systemd-user-unit-dir=/etc/systemd/user
The Meson build system
Version: 0.40.1
Source dir: /home/alex/Downloads/gamemode
Build dir: /home/alex/Downloads/gamemode/build
Build type: native build
Project name: gamemode
Native c compiler: cc (gcc 5.4.0)
Build machine cpu family: x86_64
Build machine cpu: x86_64
Found pkg-config: /usr/bin/pkg-config (0.29.1)
Native dependency libsystemd found: YES 229
Dependency threads found: YES
Library dl found: YES
Configuring config.h using configuration
Cloning into 'inih'...
remote: Counting objects: 341, done.
remote: Compressing objects: 100% (3/3), done.
remote: Total 341 (delta 0), reused 2 (delta 0), pack-reused 338
Receiving objects: 100% (341/341), 63.05 KiB | 0 bytes/s, done.
Resolving deltas: 100% (202/202), done.
Checking connectivity... done.

Executing subproject inih.

Project name: inih
Native c compiler: cc (gcc 5.4.0)
Build targets in project: 3

Subproject inih finished.
Configuring gamemoded.service using configuration
Configuring com.feralinteractive.GameMode.policy using configuration
Message: 

    Build configuration:
    ====================

    prefix:                                 /usr
    bindir:                                 /usr/bin
    datadir:                                /usr/share
    libdir:                                 /usr/lib/x86_64-linux-gnu
    libexecdir:                             /usr/libexec
    includedir:                             /usr/include
    systemd user unit directory:            /etc/systemd/user
    PolKit Action Directory:                /usr/share/polkit-1/actions

    Options:
    ========

    daemon:                                 true
    examples:                               true
    systemd:                                true

Build targets in project: 7
+ cd build
+ ninja
[2/20] Compiling c object 'daemon/gamemoded@exe/dbus_messaging.c.o'
FAILED: daemon/gamemoded@exe/dbus_messaging.c.o 
[...]

PR #5, opened 2018-04-10
Ensure to set `_GNU_SOURCE` for POSIX compliance with signals

ikeydoherty:
The systemd bus headers aren't setting a POSIX compliance level so we
define the catch-all to ensure the build doesn't error out, as seen with
issue #3. This is due to the reliance of `siginfo_t`, which requires a
minimum POSIX level of `199309L` when using glibc.

Signed-off-by: Ikey Doherty <ikey@solus-project.com>


ikeydoherty:
Ta

PR #10, opened 2018-04-10
README: added Fedora build requirements.

baxeno:
This adds prerequisites install guide to Fedora.

Signed-off-by: Bruno Thomsen <bruno.thomsen@gmail.com>


mdiluz:
Thanks, I think this will be about the limit in terms of custom dependency install commands I'll add. It could get somewhat over the top :)

baxeno:
Okay, I also think it's a nice selection of distribution families with the current list.

Issue #12, opened 2018-04-10
I can't run the daemon


leillo1975:
When I execute "sudo ./bootstrap.sh" I have the following error:
`Failed to connect to bus: Non hai tal ficheiro ou directorio`  (in English: There is no such file or directory)

You can see the process in terminal:
```
sudo ./bootstrap.sh
+ meson --prefix=/usr build -Dwith-systemd-user-unit-dir=/etc/systemd/user
Trying to run Meson on a build directory that has already been configured.
If you want to build it, just run your build command (e.g. ninja) inside the
build directory. Meson will autodetect any changes in your setup and regenerate
itself as required. Though it shouldn't be necessary, running ninja reconfigure
will force Meson to regenerate the build files.

If you want to change option values, use meson configure instead.
+ cd build
+ ninja
ninja: no work to do.
+ set +x
Install to /usr? [Yy] y
+ sudo ninja install
[0/1] Installing files.
Installing lib/libgamemode.so to /usr/lib/x86_64-linux-gnu/libgamemode.so
Installing lib/libgamemodeauto.so to /usr/lib/x86_64-linux-gnu/libgamemodeauto.so
Installing daemon/gamemoded to /usr/bin/gamemoded
Installing daemon/cpugovctl to /usr/libexec/cpugovctl
Installing gamemode_client.h to /usr/include/
Installing /home/leo/gamemode/data/gamemoded.1 to /usr/share/man/man1
Installing /home/leo/gamemode/build/data/gamemoded.service to /etc/systemd/user
Installing /home/leo/gamemode/build/data/com.feralinteractive.GameMode.policy to /usr/share/polkit-1/actions
+ set +x
Enable and run the daemon? [Yy] y
+ systemctl --user daemon-reload
Failed to connect to bus: Non hai tal ficheiro ou directorio
```
I use Ubuntu 17.10

CasperAlant:
Run without `sudo`. Only use `./bootstrap.sh`.

leillo1975:
Without sudo:
```
+ meson --prefix=/usr build -Dwith-systemd-user-unit-dir=/etc/systemd/user
Trying to run Meson on a build directory that has already been configured.
If you want to build it, just run your build command (e.g. ninja) inside the
build directory. Meson will autodetect any changes in your setup and regenerate
itself as required. Though it shouldn't be necessary, running ninja reconfigure
will force Meson to regenerate the build files.

If you want to change option values, use meson configure instead.
+ cd build
[...]

PR #13, opened 2018-04-10
Prevent running bootstrap.sh as root

CasperAlant:
This pull request prevents users from running the script as root. This will prevent issue #12 from happening.


mdiluz:
This is a sensible change, thanks for that

PR #27, opened 2018-04-12
Minor C cleanup

LeonardKoenig:
I've made some style fixes, 
 - some symbols can be made static:
    1. `set_gov_state`
    2. everything in gamemode\_client.h
 - `daemonize()` can also take a `const char*`, since `name` is only passed to `printf()` or `syslog()`
 - prevent shadowing of variables
 - use explicit `(void)` as parameter-list more consistently
 - use some more `const`.  
   Move cast to more appropriate place and document that `execv()` behaves as if args where of type `const *char` and we trust on that.
 - example: Just use main(void), which is also an acceptable ISO-C decl
 - example: Use stderr for errors

I've also thought that one could perhaps put the `typedef`s in gamemode\_client.h into an extra header to also include in client\_impl.h st. eventual changes to the interface are recorded in both there, eg:

client\_impl.h:
```c
#ifndef CLIENT_IMPL_H
#define CLIENT_IMPL_H

/* Typedefs for the functions to load */
typedef int (*_gamemode_request_start)(void);
typedef int (*_gamemode_request_end)(void);
typedef const char *(*_gamemode_error_string)(void);

#ifndef CLIENT_GAMEMODE_H /* we are not gamemode_client.h but client_impl.h */
/* forward declarations */
_gamemode_request_start gamemode_request_start;
_gamemode_request_end gamemode_request_end;
_gamemode_error_string gamemode_error_string;
#endif

#endif
````

### DIFF
--- a/daemon/cpugovctl.c
+++ b/daemon/cpugovctl.c
@@ -40,7 +40,7 @@ POSSIBILITY OF SUCH DAMAGE.
 /**
  * Sets all governors to a value
  */
-void set_gov_state(const char *value)
+static void set_gov_state(const char *value)
 {
 	char governors[MAX_GOVERNORS][MAX_GOVERNOR_LENGTH] = { { 0 } };
 	int num = fetch_governors(governors);
--- a/daemon/daemonize.c
+++ b/daemon/daemonize.c
@@ -39,7 +39,7 @@ POSSIBILITY OF SUCH DAMAGE.
 /**
  * Helper to perform standard UNIX daemonization
  */
-void daemonize(char *name)
+void daemonize(const char *name)
 {
 	/* Initial fork */
 	pid_t pid = fork();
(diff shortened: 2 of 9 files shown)


mdiluz:
Thanks,

I need to verify a few things in this - primarily the `static` additions to the functions in gamemode_client.h. But the rest is sensible enough.

For the suggestion on the split header, I like the idea, but while I agree that it'd be more obvious to see interface changes, I do want to keep the single-header approach if possible, so probably wouldn't expand to two just for neatness.

LeonardKoenig:
Hi, the `static`s should be fine, since the header is just directly included into the source code and compiled in that compilation unit, at least as far as I can gather.

Yeah, another pro of it as that you can compile with warnings that warn about missing prototypes for external functions, the idea of which is that the compiler "forces" you to either declare a forward-declare a function and let it be `extern` or make it `static`, which helps spotting functions that you either missed writing into a header-file or forgot to make `static` because they should actually have internal linkage.

One could of course just do a hack of forward-declaring them once in the client\_impl.h to shut that up (or just don't warn), it's mostly a style-choice.

PR #28, opened 2018-04-12
Integrated sd_notify to get status messages in systemd

OlliC:
Hi, 

i thought it would be cool to have status messages in systemctl's status output wether GameMode is currently active or not. systemd provides the sd_notify tool for this. The service file only needs the NotifyAccess option set so that systemd accepts notify events from that unit.

I added the message "GameMode is ready to be activated." in color Cyan when first started. 
When activated it says "GameMode is now active." in green. When inactive it says "GameMode is currently deactivated." 

Of course messages and colors are only a suggestion and can be changed. Also maybe i missed some places where the status message needs to be updated and somebody should take a look at this. 

Thanks.

### DIFF
--- a/daemon/dbus_messaging.c
+++ b/daemon/dbus_messaging.c
@@ -145,7 +145,7 @@ void game_mode_context_loop(GameModeContext *context)
 	}
 
 	LOG_MSG("Successfully initialised bus with name [%s]...\n", "com.feralinteractive.GameMode");
-	sd_notifyf(0,"STATUS=%sGameMode is ready to be activated.\n", "\x1B[1;36m");
+	sd_notifyf(0,"STATUS=%sGameMode is ready to be activated.%s\n", "\x1B[1;36m", "\x1B[0m");
 
 	/* Now loop, waiting for callbacks */
 	for (;;) {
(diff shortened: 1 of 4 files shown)


mdiluz:
I like this, so it'll go on a merge list (once I've sorted out a branching model, sorry!).

I'm also curious if other notify variables are appropriate, like say ERRNO/BUSERROR for when we fail, or STOPPING and RELOADING.

OlliC:
Great. And yes other variables can be send too. Some can be send with any unit type, but if you want to inform systemd about finished startup with READY=1 then you have to use Type=notify instead of dbus. 

See: https://www.freedesktop.org/software/systemd/man/sd_notify.html

mdiluz:
One other thing that might be good to improve on this is to keep the systemd code more isolated - to keep the goal of ensuring a non-systemd implementation is still easy to implement. If a v2 of the PR is possible with that in mind then that'd be great.

OlliC:
I will have a look at this and try to isolate it a bit more. Although i think its not a problem to run the daemon with other service managers even with sd_notify used for every status change.

mdiluz:
Yup, it appears `sd_notify` is safe to call regardless, but ideally the daemon would be able to be compiled without the systemd headers installed at all.

Obviously, this is only future-proofing, so it's not 100% required for your PR at this stage.

mdiluz:
Hi @OlliC, I've decided to accept this PR for now. I think it's only a small amount of work, when appropriate, to redesign or wrap these notify calls when a non-systemd PR comes along.

OlliC:
Great. Thank you.

PR #31, opened 2018-04-14
Added SIGTERM kill signal handling

OlliC:
Hi, 

while playing around with the `sd_notify`enabled version i have noticed that the daemon never shuts itself down cleanly when being stopped by systemd. So it never logged "Quitting by request..." and resets the systemd status message. 

Apparently systemd sends a SIGTERM signal first to request the daemon to shutdown and after a delay SIGKILL will be send. So by handling SIGTERM the daemon can cleanly shutdown.


mdiluz:
Thanks for the bug spot and fix, merged

PR #56, opened 2018-05-23
Leverage soft real time capabilities

kakra:
Kernels that support SCHED_ISO scheduling policy can give processes soft
real time support. This improves latency without compromising system
stability. See https://lwn.net/Articles/720227/.

This commit adds support for setting this policy with a safe fall back if
kernel support is lacking by just ignoring the error condition.

Additionally, it also tries to raise the nice priority of the game to -4
to give it a slight IO and CPU priority over other background processes.
This needs PAM adjustments to allow users raising priority to certain
levels. If it doesn't work, the fall back strategy is also ignoring the
error condition. See /etc/security/limits.conf.

Kernels that currently support SCHED_ISO include kernels with Con
Kolivas MuQSS patchset (likely the CK patchset). This patchset is
generally recommended for desktop machines but usually not found in
standard distribution kernels due to lack of widespread stability tests.


mdiluz:
This is cool, the two parts could also be plugged into the config system so they can be turned on/off, and perhaps the niceness level customized.

kakra:
Yes, this should be configurable. Because there's one issue that could hit some games: Games that use busy looping with too few cores in the system may lead to priority inversion, although I didn't yet experience such a problem.

Do you want to merge first or should I try to add configuration first?

I could also try to extend this to support CPU affinity tho I don't know how to properly auto-configure this for "CPU known to work better with it"... It was mentioned in some issue here. And owning an Intel system I don't see much point of adding it.

kakra:
BTW: Most software seems to agree that adjusting nice levels by -4, 4 or 5 can be hard-coded, e.g. Chrome runs some of its sub processes nice+5 while some distributions adjust the X server to nice-4. It correlates  with IO priorities:

`io_nice = (cpu_nice + 20) / 5`
(https://www.kernel.org/doc/Documentation/block/ioprio.txt)

So, -4 would adjust IO priority to be a little higher because (20-4)/5 = 16/5 = BE3 while the default is 20/5 = BE4. The other way around, +4 wouldn't adjust IO priority (still BE4) while +5 does (BE5).

Thus, we should maybe skip nice adjusting if the user manually already starts the process reniced.

kakra:
Okay, here's configuration. I thought "why wait?", so I've just accepted the challenge. Please have a look.

aejsmith:
Could you make the reniceing configurable as well? Generally I'd prefer everything to be configurable, as it makes it easier to test the performance impact of individual changes.

Otherwise, looks good to me. @mdiluz OK with you?

kakra:
I'm looking into it, the test scenario makes sense. I'll also document the priority inversion caveats then. Thanks for considering.

kakra:
@aejsmith I've implemented configuration for the renice value. Please review. From my point of view it can be merged now.

aejsmith:
Sorry for the delay. Looks good to me, I've merged it. Thanks!

Issue #61, opened 2018-06-28
Fedora package related issues (lib versioning)


gicmo:
Hi,
I work in Red Hat's Desktop Hardware Enablement team and are currently working on packaging gamemode for Fedora (see [fedora-workstation issue 61](https://pagure.io/fedora-workstation/issue/61), review request [rhbz#1596293](https://bugzilla.redhat.com/show_bug.cgi?id=1596293)).
After doing the initial packaging I have a couple of questions/issues:

* Issues:
    - [ ] location of the libraries (currently `/usr/lib64`)
  They are unversioned which normally is a big no-no: "In cases where upstream ships unversioned .so library (so this is not needed for plugins, drivers, etc.), the packager MUST try to convince upstream to start versioning it."  ([see "Downstream .so name versioning"](https://fedoraproject.org/wiki/Packaging:Guidelines#Shared_Libraries)). Maybe that is ok to not have them versioned, because they are internal, i.e. get dlopen'ed by the inlined functions from the inlined functions in the header, but then we should install them in a private location (i.e. `/usr/lib64/gamemode/` like we did for renderdoc). OTOH, umockdev also has a private, ld-preload only library and that is versioned: `/usr/lib64/libumockdev-preload.so.0` & `libumockdev-preload.so.0.0.0`.

* Questions:
    - [ ] location of the gamemoded daemon (currently `/usr/bin`)
  Any reason why it is not in `/usr/libexec/`?

    - [ ] man page section for `gamemoded` (1 vs 8)
  It is a daemon, so should be in section 8?

More out of curiosity, why a session daemon and not a system one?

aejsmith:
Thanks for working on packaging this. I've merged your pull request.

I don't think there's any particular reason for /usr/bin vs /usr/libexec. Is Fedora's preferred location for daemons /usr/libexec?

You're right on the man page section. I'll move it.

There's a couple of reasons for using a session daemon:
- Means that the whole daemon doesn't run with elevated privileges, just the parts that need it (cpugovctl).
- Allows configuration on a per-user basis - one of the locations it attempts to read the configuration from is the the user's home directory.

gicmo:
Thanks for the quick review of the patches!

[...]

PR #63, opened 2018-07-02
Introduce libtool like library versioning

gicmo:
Addressing the problem brought up in #61 that for packaging in Fedora (and probably others, e.g. [debian](https://www.debian.org/doc/debian-policy/#document-ch-sharedlibs)) that shipped libraries must be versioned, we do so via libtool like versioning scheme (`current.revision.age`), which for now is just `0.0.0`. The client header then uses the versioned library, i.e. `libgamemode.so.0`.
Additionally provide package config files to ease development.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,13 +57,13 @@ This will also satisfy the build requirement `inih` by pulling it in as a git su
 ## Requesting GameMode
 
 ### Users
-After installing `libgamemodeauto.so` simply preload it into the game:
+After installing `libgamemodeauto.so.0` simply preload it into the game:
 ```bash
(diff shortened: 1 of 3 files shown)


fuckyourselfup:
If merged this commit, there should be symlinks like these ("x" is number of version, revision, etc):

libgamemode.so -> libgamemode.so.0.x.x
libgamemode.so.0 -> libgamemode.so.0.x.x
libgamemode.so.0.x -> libgamemode.so.0.x.x

aejsmith:
Thanks for these changes! They look good to me - symlinks do already get created as per the previous comment, which keeps backwards compatibility with games built against previous headers.

PR #71, opened 2018-07-30
Bugfixes and Hardening

mgerstner:
These are a couple of bugfixes and improvements I came up with during a security review for openSUSE (see [openSUSE bug](https://bugzilla.suse.com/show_bug.cgi?id=1093979)).


aejsmith:
Thanks for this! All looks good to me, merging.

PR #79, opened 2018-09-29
Identify original wine exe by PID

kakra:
This PR adds a series of commits to detect the original game exe name from the PID if GameMode is used to run wine programs. Without this, we are not able to blacklist specific games or executables because every wine PID identifies as `wine-preloader` or `wine64-preloader`.

Please review and share your ideas. ~~I still have some fixes and todos to work on. But if you think this isn't worth inclusion or should go another route, it's good to have feedback early.~~

### DIFF
--- a/daemon/gamemode.c
+++ b/daemon/gamemode.c
@@ -70,6 +70,11 @@ POSSIBILITY OF SUCH DAMAGE.
  */
 #define CLAMP(lbound, ubound, value) MIN(MIN(lbound, ubound), MAX(MAX(lbound, ubound), value))
 
+/* Little helper to safely print into a buffer, returns a pointer into the buffer
+ */
+#define buffered_snprintf(b, s, ...)                                                               \
+	(snprintf(b, sizeof(b), s, __VA_ARGS__) < (ssize_t)sizeof(b) ? b : NULL)
+
 /* Little helper to safely print into a buffer, returns a newly allocated string
  */
 #define safe_snprintf(b, s, ...)                                                                   \
@@ -753,6 +758,11 @@ static char *game_mode_resolve_wine_preloader(pid_t pid)
 	char *proc_path = NULL, *wine_exe = NULL, *wineprefix = NULL;
 	int proc_fd = -1;
 
+	/* We could use the buffered_snprintf() helper here but it may potentially
+	 * overwrite proc_path when the buffer is re-used later and usage of
+	 * proc_path has not been discarded yet (i.e., it's used in the fail path).
+	 * Let's not introduce this non-obvious pitfall.
+	 */
 	if (!(proc_path = safe_snprintf(buffer, "/proc/%d", pid)))
 		goto fail;
 
@@ -818,10 +828,9 @@ static char *game_mode_resolve_wine_preloader(pid_t pid)
 	wine_exe[0] = (char)tolower(wine_exe[0]);
 
 	/* Convert relative wine exe path to full unix path */
-	char *wine_path = safe_snprintf(buffer, "%s/dosdevices/%s", wineprefix, wine_exe);
+	char *wine_path = buffered_snprintf(buffer, "%s/dosdevices/%s", wineprefix, wine_exe);
 	free(wine_exe);
 	wine_exe = wine_path ? realpath(wine_path, NULL) : NULL;
-	free(wine_path);
 
 	/* Fine? Successo? Fortuna! */
 	if (wine_exe)
@@ -861,12 +870,11 @@ static char *game_mode_context_find_exe(pid_t pid)
 	char buffer[PATH_MAX];
 	char *proc_path = NULL, *wine_exe = NULL;
 
-	if (!(proc_path = safe_snprintf(buffer, "/proc/%d/exe", pid)))
+	if (!(proc_path = buffered_snprintf(buffer, "/proc/%d/exe", pid)))
 		goto fail;
 
 	/* Allocate the realpath if possible */
 	char *exe = realpath(proc_path, NULL);
-	free(proc_path);
 	if (!exe)
 		goto fail;
 
--- a/daemon/gamemode.c
+++ b/daemon/gamemode.c
@@ -454,14 +454,17 @@ bool game_mode_context_register(GameModeContext *self, pid_t client)
 		fputs("OOM\n", stderr);
 		return false;
 	}
-	cl->executable = game_mode_context_find_exe(client);
-	if (!cl->executable)
-		goto error_cleanup;
 
+	/* Check the PID first to spare a potentially expensive lookup for the exe */
 	if (game_mode_context_has_client(self, client)) {
 		LOG_ERROR("Addition requested for already known process [%d]\n", client);
 		goto error_cleanup;
 	}
+
+	/* Lookup the executable */
+	cl->executable = game_mode_context_find_exe(client);
+	if (!cl->executable)
+		goto error_cleanup;
 
 	/* Check our blacklist and whitelist */
 	if (!config_get_client_whitelisted(self->config, cl->executable)) {
--- a/daemon/gamemode.c
+++ b/daemon/gamemode.c
@@ -433,6 +433,9 @@ static void game_mode_context_auto_expire(GameModeContext *self)
 			pthread_rwlock_unlock(&self->rwlock);
 			break;
 		}
+
+		if (game_mode_context_num_clients(self) == 0)
+			LOG_MSG("Properly cleaned up all expired games.\n");
 	}
 }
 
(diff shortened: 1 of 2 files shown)


daemon/gamemode.c
```diff
@@ -604,6 +605,78 @@ GameModeContext *game_mode_context_instance()
...
 
+/**
+ * Attempt to resolve the exe for wine-preloader.
+ * This function is used if game_mode_context_find_exe() identified the
+ * process as wine-preloader.
+ */
+static char *game_mode_resolve_wine_preloader(pid_t pid)
+{
+	static char proc_path[PATH_MAX] = { 0 };
+	int fd = 0, proc_fd = 0;
```
aejsmith: 0 is a valid FD, even if it is usually taken by stdin. Better to initialise to -1 and change the cleanup code accordingly.

daemon/gamemode.c
```diff
@@ -604,6 +605,78 @@ GameModeContext *game_mode_context_instance()
...
 }
 
+/**
+ * Attempt to resolve the exe for wine-preloader.
+ * This function is used if game_mode_context_find_exe() identified the
+ * process as wine-preloader.
+ */
+static char *game_mode_resolve_wine_preloader(pid_t pid)
+{
+	static char proc_path[PATH_MAX] = { 0 };
```
aejsmith: Any reason for static here (and later on)? Makes this function non-thread-safe. Stack size could be a concern but I don't think 2 4K buffers is too much issue given that default stack sizes should be significantly larger.

aejsmith:
This seems like a useful thing to do to me. I don't know anything about what Wine processes look like so I can't really comment on whether this is technically correct, but the code looks good aside from the comments I've added, and I assume you've tested it.

kakra:
@aejsmith You added comments? I don't see any... Did you use the review function of Github?

aejsmith:
Oops, didn't hit the submit button.

daemon/gamemode.c
kakra: Good point, will do.

daemon/gamemode.c
kakra: I think I copied that from somewhere else and wondered about thread-safety, too. I'll change that. What about using `malloc()` to not pullute the stack? I think an allocation is not a performance problem here, this only loads at process startup time. Or I try to move this logic up into the config parser.

kakra: @aejsmith I actually copied it from `game_mode_find_exe()` which has this static line. Since my function is always only called from that said function, it should have the same semantics, shouldn't it? But I can fix that function, too, if thread-safety is a concern here. Are these functions really called from different threads?

daemon/gamemode.c
kakra: Fixed.

daemon/gamemode.c
kakra: I rebased a fix you can cherry-pick from.

kakra:
It's not entirely tested yet but feel free to review. I'd like to give it a view days in my test bed first. At least valgrind is clean... C strings are really a, well, bottom itch? PITA? :-)

kakra:
Okay, some more fixups... We need to explicitly ignore some detection scenarios when running under wine because otherwise processes are mis-detected or end up as game "(null)" in the client list. I've rebased and optimized some minor issues.

Wine processes seem to fork themselves multiple times, the PID doesn't change and then GameMode would think the client is already in the list but doesn't know its final name (which is unfortunate for proper blacklisting/scripting/etc). This seems to be fixed now.

Here's a log for both ROTTR and SOTTR:
https://gist.github.com/kakra/ac1f8a674951f7ced1f862a96e5bdfe2

PS: I hope you Feral guys are working on a native port of SOTTR. :-)

daemon/gamemode.c
aejsmith: I don't think it can be used by multiple threads at the same time right now, just don't think it's a good idea to unnecessarily make the restriction that it can't be.

Hadn't noticed `game_mode_find_exe` had the same, so yes, that should be fixed too.

daemon/gamemode.c
```diff
@@ -59,6 +59,11 @@ POSSIBILITY OF SUCH DAMAGE.
  */
 #define NICE_DEFAULT_PRIORITY -4
 
+/* Little helper to safely print into a buffer, returns a newly allocated string
+ */
+#define safe_snprintf(b, s, ...)                                                                   \
+	(snprintf(b, sizeof(b), s, __VA_ARGS__) < (ssize_t)sizeof(b) ? strndup(b, sizeof(b)) : NULL)
+
```
aejsmith: Don't see why this needs to allocate a new string as well seeing as we've already supplied the buffer? Could just return b or null as an indicator of success.

aejsmith: Also, this commit is in the wrong order - "Add function to resolve the wine preloader executable" is before it and depends on it.

daemon/gamemode.c
```diff
@@ -399,6 +399,8 @@ bool game_mode_context_register(GameModeContext *self, pid_t client)
 		return false;
 	}
 	cl->executable = game_mode_context_find_exe(client);
+	if (!cl->executable)
+		return false;
```
aejsmith: Should be `goto error_cleanup`, this'll leak `client`.

aejsmith:
Aside from the comments I've added, LGTM. Let me know once you're happy with it after testing.

daemon/gamemode.c
kakra: The idea was having a reuseable and disposable buffer explicitely, and having a string as a result I can work with without having to think of `strdup()`'ing it extra. I actually had it the way you suggested it first and it was a complete mess. I had multiple buffers per stack or had to `strdup()` all the time just to make sure I would not overwrite my string in the next buffer re-use. There's maybe one code location now where we do a `strdup()` too much now but I don't think it's worth microoptimizing this. If you want to re-use the helper in other parts of the code, I could think of making two versions: One with and one without the `strdup()`. But currently it makes my code more readable and shorter.

Order: Yes, I noticed that, too. I checked in the functions one after each other during a longer edit session, then rebased afterwards to put it into correct commit order. Github insists on displaying this in order of commit date, instead of parent commit order (and even that gets messed up with partial rebases). Use `tig` on the command line to see it correctly. :-)

daemon/gamemode.c
kakra: It's actually fixed, it's one of the top commits.

daemon/gamemode.c
kakra: PS: In compare view, is actually in correct order - https://github.com/FeralInteractive/gamemode/compare/master...kakra:identify-wine-exe

I think it's in a different order here for better overview of what changed after a rebase, to easy up incremental reviews.

kakra:
If you want to merge the iopriority branch first, this one needs a rebase to be merged without conflicts. The conflicts are a bit messy to resolve so I would do it for you. From what I've tested so far it seems to be working very well and the logs look clean. After final rebase I'd do another round of code cleanup (there are one or two locations I don't like visually) and then check `valgrind` again.

daemon/gamemode.c
kakra: When I do the final code cleanup, I will look into making two versions of the helper and see if it optimizes things. But I think the helper is fine in view of how it is supposed to be used. Do you agree? A non-`strdup()` version clearly makes sense where the buffer is a one-time use only. I'll look into that but otoh I hope that you agree with me that a `strdup()` version is worth having. :-)

daemon/gamemode.c
kakra: Okay, I'll look into it. This is also old code where we found problems. :-)

kakra: Scrap that... It's new code by me. But what about the other above, the OOM thing?

kakra: Scrap that again... Good morning to me. :-)

kakra: I see potential for cleanup here anyways, will push with the next rebase.

daemon/gamemode.c
kakra: @aejsmith I've added a non-`strdup()` version of the helper. But when used improperly, it may introduce more headache than it optimizes... I'm not sure if we want to keep it in. You decide...

daemon/gamemode.c
kakra: I cleaned up the error path and general handling in the register routine. You may want to take a look.

kakra:
The new rebase of the patch series adds some cleanup in the code, some cleaned up error paths, and also introduces some better logging because testing showed that wine processes may show up with some concerning logs. We don't want to worry any one unnecessary, do we? ;-)

kakra:
If you're happy with that, I'm happy, too. :-)

aejsmith:
OK, all looks good. As you said it looks like it needs rebasing now since I've just merged the I/O scheduling branch. If you do that, could you squash the last two logging commits as well?

kakra:
Rebased and squashed the last two commits (which makes sense in combination with the other logging improvements).

daemon/gamemode.c
kakra: I understand your answer below as "yes, keep it in" and resolve this conversation. :-)

aejsmith:
Merged, thanks!

PR #81, opened 2018-09-30
Io priorities

kakra:
This PR allows changing the IO priority of the client to a value specified in the configuration file. That can possibly reduce lags or latency when a game has to load assets on demand and you have background IO activity running (or other concurrent IO).

First, we add some helper functions, the second commit adds the configuration options and the final commit adds the functionality. We also need to define the `syscall()` interface for `ioprio_set()` because glibc provides no wrapper for it.

This will come into play better when the PR for identifying wine processes will be merged, as it allows fine grained control of only prioritizing the game and not various other processes that spawn with SteamPlay/Proton.


data/gamemoded.8
```diff
@@ -110,6 +110,12 @@ softrealtime=auto
 ; the value will be negated and applied as a nice value
 renice = 4
 
+; By default, GameMode adjusts the iopriority of clients to BE/0, you can put any value
+; between 0 and 7 here (with 0 being highest priority), or one of the special values
+; "off" (to disable) or "reset" (to restore Linxu default bahavior based on CPU priority),
```
bunder2015: Linux

example/gamemode.ini
```diff
@@ -15,6 +15,12 @@ softrealtime=auto
 ; the value will be negated and applied as a nice value
 renice = 4
 
+; By default, GameMode adjusts the iopriority of clients to BE/0, you can put any value
+; between 0 and 7 here (with 0 being highest priority), or one of the special values
+; "off" (to disable) or "reset" (to restore Linxu default bahavior based on CPU priority),
```
bunder2015: Linux

data/gamemoded.8
bunder2015: behaviour

example/gamemode.ini
bunder2015: behaviour

data/gamemoded.8
i-need-to-tell-you-something: > behaviour

or behavior

kakra: I'm used to american English better, so I use "behavior". But I'm fixing this...

i-need-to-tell-you-something: Yes, i agree, but you spelled "behavior" as "bahavior". ;)

kakra: What!? Oh noes! Will fix... :-)

aejsmith:
Nice idea. Code looks good to me.

Do users require any privileges to change the I/O priority beyond what would usually be the default?

kakra:
Users are free to choose from any priority within the best-effort class (which is why I didn't make this configurable). With modern kernels (post 2.6), it is also possible to set own processes to idle priority but this makes no sense either, thus not configurable.

The other case would be realtime IO priority. But I think this shouldn't be used at all for a game because it could totally dominate all IO (realtime class will be always served before anything else), and could make it difficult to kill a miss-behaving game. So, this is also covered by not being able to configure the class. Of course, brave users could still use scripts to set this and configure GameMode to not touch the IO instead. This option is implemented.

Issue #68, opened 2018-07-21
Setting process to SCHED_ISO fails with error 1


terencode:
**Describe the bug**
My kernel provides SCHED_ISO and gamemode detects it but fails to apply it: 
>"Setting client [22578] to SCHED_ISO failed with error 1, ignoring"

**To Reproduce**
Steps used to reproduce the behavior:
1. Launch gamemod with a kernel providing SCHED_ISO
2. Run any game
3. View output of journalctl /usr/bin/gamemoded
4. See error "Setting client [PID] to SCHED_ISO failed with error 1, ignoring"

**Expected behavior**
Gamemode should be able to set the process to SCHED_ISO
**System Info:**
 - OS and version: ArchLinux 4.17.6-1-lqx
 - GameMode Version 1.2

**Additional context**
I would think this is a permission/configuration problem.

VanackSabbadium:
I've got the same problem, and i also add that GameMode can't renice processes.

`Renicing client [3499] failed with error 13, ignoring.`

aejsmith:
Renicing will likely fail by default due to permissions. You need to configure `/etc/security/limits.conf` to allow regular users to raise the priority.

I'm not sure if you need to do similar with `SCHED_ISO` - @kakra do you know if there's anything you need to do permissions-wise for it to work?

kakra:
`SCHED_ISO` is to my best knowledge not privileged because it's safe to use (it does not dominate or destabilize the system). Still, errno = 1 means `EPERM`. I don't think this is a permission error by the user lacking permission (this would result in errno = 13, `EACCES`) but the operation is simply not permitted by system restrictions. This usually indicates a missing feature or unavailable due to configuration.

@terencode Did you compile the kernel with `CONFIG_SCHED_MUQSS=y`? It is not sufficient to just patch the kernel with the CK patchset and then recompile. You must reconfigure to enable MuQSS which is the feature that actually makes `SCHED_ISO` available.

BTW @VanackSabbadium You probably need to restart your session after adjusting pam limits. Otherwise, the limits aren't loaded into the current session and you won't be able to use them.

kakra:
@terencode Can you run `schedtool -I -e /path/to/game/exe` and see if it complains? It should work out of the box if your kernel has the required configuration and schedtool is new enough.

kakra:
[...]

PR #84, opened 2018-10-03
gamemode-scheduler: Improve logging

kakra:
During finding the cause of an issue, we discovered bad interactions
with other resource management daemon that may be running at the same
time and may interfere with GameMode's ability to properly set
scheduling parameters of the process. Usually, this results in EPERM
(instead of EINVAL for lacking kernel support). It also means that the
other daemon is leaking wrong scheduling settings into the games with
negative performance impacts (quite the opposite of what the user wanted
to achieve). So we should give some valuable hints.

This commit improves logging by detecting the error situation and giving
context-based hints on what to do next. To keep logging verbosity low,
the hint is only logged once per error code.

Closes: https://github.com/FeralInteractive/gamemode/issues/68
Signed-off-by: Kai Krakow <kai@kaishome.de>

### DIFF
--- a/daemon/gamemode.c
+++ b/daemon/gamemode.c
@@ -184,15 +184,15 @@ void game_mode_context_destroy(GameModeContext *self)
  */
 static void game_mode_apply_scheduler(GameModeContext *self, pid_t client)
 {
-	LOG_MSG("Setting scheduling policies...\n");
-
 	/*
 	 * read configuration "renice" (1..20)
 	 */
 	long int renice = 0;
 	config_get_renice_value(self->config, &renice);
 	if ((renice < 1) || (renice > 20)) {
-		LOG_ERROR("Renice value [%ld] defaulted to [%d].\n", renice, -NICE_DEFAULT_PRIORITY);
+		LOG_ERROR("Invalid renice value '%ld' reset to default: %d.\n",
+		          renice,
+		          -NICE_DEFAULT_PRIORITY);
 		renice = NICE_DEFAULT_PRIORITY;
 	} else {
 		renice = -renice;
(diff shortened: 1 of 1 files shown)


daemon/gamemode.c
```diff
@@ -225,14 +226,45 @@ static void game_mode_apply_scheduler(GameModeContext *self, pid_t client)
...
+			const char *additional_message = "";
+			switch (errno) {
+			case EPERM: {
+				static int once = 0;
+				if (once++)
+					break;
+				additional_message =
+				    "    -- The error indicates that you may be running a resource management\n"
+				    "    -- daemon managing your game launcher and it leaks lower scheduling\n"
+				    "    -- classes into the games. This is likely a bug in the managment daemon\n"
```
aejsmith: s/managment/management/

daemon/gamemode.c
```diff
@@ -225,14 +226,45 @@ static void game_mode_apply_scheduler(GameModeContext *self, pid_t client)
...
-			    "kernel may not support this).\n",
+			    "Failed to set client [%d] into SCHED_ISO mode, ignoring error condition: %s\n%s",
 			    client,
-			    errno);
+			    strerror(errno),
+			    additional_message);
 		}
 	} else {
-		LOG_ERROR("Not using softrealtime, setting is '%s'.\n", softrealtime);
+		LOG_ERROR("Skipping to set client [%d] into SCHED_ISO mode: softrealtime setting is '%s'\n",
```
aejsmith: "Skipping setting"

terencode:
For Ananicy, this should be fixed as of https://github.com/Nefelim4ag/Ananicy/commit/645f2f5f87748a4ad3ac60a25de7437605ec4b52

daemon/gamemode.c
kakra: Argh, I should make an auto-correction for that word. I know how to spell it correctly but always type it wrong. ;-)

daemon/gamemode.c
kakra: Ah yes the fun of not being the native speaker. I corrected that into "Skipped setting" to be inline with the message above.

PR #86, opened 2018-10-08
Tidy up the code and refactor into modules

kakra:
The previous PRs have shown that the code base grows too big to keep everything in one file. Thus, let's refactor it and break out functionality into separate C files. This returns `gamemode.c` almost back into its original appearance. It should result in fewer rebase and merge conflicts in the future.

This also cleans up the logging code a bit by adding some macros for easy hinting.


aejsmith:
Looks good to me, thanks for doing this.

PR #91, opened 2018-12-02
Configurable prefix path

suvayu:
This PR makes the prefix path in `bootstrap.sh` configurable by setting a temporary variable during invocation.

    $ prefix=/alternate/path ./bootstrap.sh

In a separate commit, the default is also changed to `/usr/local` as it is customary to install external software under `/usr/local`.  If the change of default is not acceptable, please feel free to revert it (or let me know, and I'll revert).

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -19,15 +19,18 @@ if [ ! -f "/sys/devices/system/cpu/cpu0/cpufreq/scaling_governor" ]; then
 	[[ $REPLY =~ ^[Yy]$ ]]
 fi
 
+# accept a prefix value as: prefix=/path ./bootstrap.sh
+: ${prefix:=/usr}
+
 # Echo the rest so it's obvious
 set -x
-meson --prefix=/usr build -Dwith-systemd-user-unit-dir=/etc/systemd/user
+meson --prefix=$prefix build -Dwith-systemd-user-unit-dir=/etc/systemd/user
 cd build
 ninja
 
 # Verify user wants to install
 set +x
-read -p "Install to /usr? [Yy] " -r
+read -p "Install to $prefix? [Yy] " -r
 [[ $REPLY =~ ^[Yy]$ ]]
 set -x
 


aejsmith:
If installed to `/usr/local` things don't work properly for a couple of reasons:
- `/usr/local/lib(64)` isn't searched by default for libraries, at least on Fedora, so without manual changes to `ld.so.conf`, then GameMode requests from games will fail since they can't open the library.
- polkit doesn't look in `/usr/local/share/polkit-1/actions`, so the policy for that won't be found and any GameMode requests will result in an authentication prompt.

I think the default behaviour from `bootstrap.sh` should give a working installation without any additional configuration, so could you please drop that change? The configurable prefix commit, however, is fine.

suvayu:
Hmm, I have it installed in `/usr/local` and it seems to be working fine (unless it has been silently failing, and I haven't noticed).  Anyway, I have reverted that commit.

Cheers,

aejsmith:
Merged, thanks.

Could be distro-specific whether or not things work when installed in `/usr/local`. I'd just given it a go on my Fedora 29 install and found that it did not.

kakra:
Some distributions define `ld.so.conf` to include `/usr/local/lib{32,64,}`, some don't. Users are expected to adjust this file if they plan on installing lib files to non-dist-default locations. Maybe that should be documented...

suvayu:
Well, I'm on F28, and my `ld.so.conf` doesn't do anything special.

    # cat /etc/ld.so.conf
    include ld.so.conf.d/*.conf

The files in the conf directory just include library specific directories.

kakra:
As most distributions do... It's in one of those include files then. But `ld.so.conf` is the master file so my statement is still correct wrt to the distribution defining defaults via this file. What's in the `.d` directory is distributions specific, it makes no sense to mention a specific file name here.

Finally, everything should end up in `ld.so.cache` anyways but that's a binary cache file that needs to be regenerated when you updated the config (`man ldconfig`).

PR #93, opened 2019-01-23
Enable and start service in bootstrap.sh

lwindolf:
As systemd service installation is not trivial I believe many non-power users will benefit from bootstrap handling service activation. Please consider merging this!

And thanks a lot for gamemode!!!

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -38,3 +38,13 @@ sudo ninja install
 
 # Reload systemd configuration so that it picks up the new service.
 systemctl --user daemon-reload
+
+# Verify user wants to activate now
+set +x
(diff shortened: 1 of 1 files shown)


kakra:
You don't need to enable or start the service, it is dbus activated... You should run just `systemctl --user daemon-reload` and be done with it.

aejsmith:
@kakra is correct, service activation used to be done by bootstrap.sh but was removed after we switched to DBus activation. bootstrap.sh already does `daemon-reload` as well.

Do you have problems without it being manually activated?

lwindolf:
I see. Retested and DBUS activation works as described!

Withdrawing the PR. Thanks!

PR #95, opened 2019-01-26
Add integration tests

mdiluz:
This adds the `-t` argument to `gamemoded` whereby it will perform some tests on the currently installed (and ideally running) daemon.

For now this simply verifies a few key things:
- The daemon responds correctly when we request to start and end
- The daemon correctly responds when two clients are connected
- The `-r` argument correctly requests gamemode and waits for SIGINT before cleaning it up correctly as well.

This does not test any of the gamemode functionality actually correctly applies itself to the system. Yet.

Other than verifying other PRs don't break core functionality, this should be useful for prelim support when making sure the main gamemode daemon is installed and responding correctly, we can ask "What's the output of `gamemoded -t`?".

### DIFF
--- a/daemon/gamemode-tests.c
+++ b/daemon/gamemode-tests.c
@@ -47,7 +47,7 @@ static int verify_gamemode_initial(void)
 {
 	int status = 0;
 
-	if ((status = gamemode_query_status()) != 0) {
+	if ((status = gamemode_query_status()) != 0 && status != -1) {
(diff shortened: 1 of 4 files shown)


mdiluz:
Good point, I'll clean that up

mdiluz:
There we go, cleaned up both 👍

aejsmith:
Looks good, merged.

PR #97, opened 2019-01-28
Have the full bootstrap script run in the CI

mdiluz:
Travis now has `Xenial`! 🎉 

Use this to install the dependencies needed to run `bootstrap.sh` to verify that the full compile, install, and script itself work.

It doesn't look like a travis docker has dbus activated however, so we can't run `-t` yet, will need to consider options for that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,12 @@
 language: c
 compiler: gcc
 sudo: false
+dist: xenial
 
 addons:
   apt:
     packages:
-      - clang-format-3.4
+      - clang-format
 
 script:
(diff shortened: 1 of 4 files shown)


mdiluz:
7ae8dee removes the additional define

mdiluz:
@aejsmith Look like Github "did a weird" here. Maybe we acted at the same time?

aejsmith:
I merged it while it was still doing the CI run, but master branch seems fine to me?

mdiluz:
Yeah seems so, thanks!

PR #98, opened 2019-01-28
Arch users prefer installing from AUR.

yurikoles:


esmithferal:
As per Marc DL's comment, this is already documented in the Read Me so no changes are required.

PR #105, opened 2019-02-21
Detect the GPU vendor automatically

mdiluz:
Remove the `gpu_vendor` config variable, this can be auto-detected from the device info in the sysfs.

Also, refactor the testing a little to reflect this and use the `GameModeGPUInfo`  type perhaps more correctly.

### DIFF
--- a/example/gamemode.ini
+++ b/example/gamemode.ini
@@ -43,7 +43,7 @@ inhibit_screensaver=1
 ; Setting this to the keyphrase "accept-responsibility" will allow gamemode to apply GPU optimisations such as overclocks
 ;apply_gpu_optimisations=0
 
-; The device number on the system (usually 0)
+; The DRM device number on the system (usually 0), ie. the number in /sys/class/drm/card0/
 ;gpu_device=0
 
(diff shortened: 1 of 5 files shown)


example/gamemode.ini
```diff
@@ -43,9 +43,7 @@ inhibit_screensaver=1
 ; Setting this to the keyphrase "accept-responsibility" will allow gamemode to apply GPU optimisations such as overclocks
 ;apply_gpu_optimisations=0
 
-; You must set these to tell gamemode the Vendor of your graphics card, as well it's device number on the system (usually 0)
-; Vendor must be one of 0x10de (NVIDIA), 0x1002 (AMD) or 0x8086 (Intel)
-;gpu_vendor=0x0000
+; The device number on the system (usually 0)
```
aejsmith: Still think this is a bit unclear - "DRM device number" would be better.

mdiluz: Hmm yeah, I guess I could also put an example with the dev path

mdiluz: bfe0903 tries to make this better :)

aejsmith: Thanks, merged.

PR #106, opened 2019-02-21
Add "supervisor" support

mdiluz:
This adds support for a "supervisor" type process to request gamemode for other processes.

**External**
- `gamemode_request_start_for` to request gamemode starts for another process
- `gamemode_request_end_for` to request gamemode ends for another process
- Bugfix: when non-required functions (like the above) don't exist, we now safely fail, instead of outright bailing, see 852f096

These two client functions are a simple noop when an older daemon lib is installed.

**Config**
- `supervisor` config section for the below settings
- `require_supervisor` can be used to restrict all gamemode registrations to just supervisor functions
- `supervisor_whitelist` to whitelist specific supervisor programs (if this exists, anything not matching is rejected)
- `supervisor_blacklist` to blacklist specific supervisor programs (anything matching is rejected)

**Internal**
- `config_string_list_contains` checks if a string list contains a substring match
- dbus methods `RegisterGameByPID` and `UnregisterGameByPID` to assist the new client APIs

**Tests**
New test coverage for supervisor functionality. Pass looks like this locally for me:
```
:: Supervisor tests
:: Passed
```
Note: When `require_supervisor` is set the current tests will bail out, for now. A small test refactor will be needed to get that working.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,6 +148,18 @@ Developers can integrate the request directly into an app. Note that none of the
 
 Or, distribute `libgamemodeauto.so` and either add `-lgamemodeauto` to your linker arguments, or add it to an LD\_PRELOAD in a launch script.
 
+### Supervisor support
+Developers can also create apps that manage GameMode on the system, for other processes:
+
+```C
+#include "gamemode_client.h"
+
+	gamemode_request_start_for(gamePID);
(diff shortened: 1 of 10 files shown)


aejsmith:
LGTM. Do you have a specific use case in mind for this?

PR #108, opened 2019-03-07
Assorted small bug fixes

mdiluz:
I've had these sitting for a bit. Fixes include:

* Implement timeout for external script commands, controlled by `script_timeout`
* Add output for failed external commands
* Run custom scripts last always so the internal optimizations are always applied and removed timely
* Tests now wait for a bit before running, in case a process still has gamemode and will be reaped
* Log the actual dbus errors in the client to help debugging
* Make the governor test actually check the old governor first


aejsmith:
Merged, plus I've done a few extra fixes on top of this (472bdcc2a38f88f2ef583a9129e35e617e0b50ca, 0e06e68100fdc1000ecc930a847012f9154fdabd, b7a6b1cace3b47ac0d0526c25b1742a159ac59ac).

Issue #45, opened 2018-05-02
Missing DISPLAY environment variable in systemd user service after first start


OlliC:
Hi, 

i have started using the new custom user script functionality to put my Nvidia GPU into performance mode. I am using this config for that: 

```
[custom]
; Custom scripts (executed using the shell) when gamemode starts and ends
start=notify-send "GameMode started"
    nvidia-settings -a '[gpu:0]/GPUPowerMizerMode=1'

end=notify-send "GameMode ended"
    nvidia-settings -a '[gpu:0]/GPUPowerMizerMode=0'
```

The problem with this is sometimes nvidia-settings fails to run because it needs the DISPLAY enviroment variable (+ XAUTHORITY probably), which is missing when the service first gets started. 

I can see with `systemctl --user show-environment` that the variables are there. But when i look at the actual environment from the gamemoded process with `tr '\0' '\n' < /proc/<pid>/environ` they are missing. But after a service restart the variables are there. I suspect the problem is that the service is run too early before the X server had time to execute `/etc/X11/xinit/xinitrc.d/50-systemd-user.sh` (See [1]) which source these variables into systemds user daemon. 

By using a small delay with `ExecStartPre=/usr/bin/sleep 5` the problem is gone, but there must be a better solution to this. I have this problem on Fedora and Arch Linux so it probably affects all distros. 

[1] https://wiki.archlinux.org/index.php/Systemd/User#DISPLAY_and_XAUTHORITY

mdiluz:
Cheers @OlliC this is a good report.

So from what I'm aware it's complicated to get this right, see: [this superuser post](https://superuser.com/questions/759759/writing-a-service-that-depends-on-xorg) for discussions. Would be open to suggestions about clean ways, as it's unfortunate you can't just specify `Wants=xorg.service`!

OlliC:
My current workaround is to restart the service with Gnomes autostart feature with this desktop file: 

```
$ cat .local/share/applications/gamemoded-restart.desktop 
[Desktop Entry]
Name=Restart gamemoded user service
Exec=systemctl --user restart gamemoded.service
Type=Application
Terminal=false
```

This works quite well and is probably better than sleep for 5 seconds, because the time to startup the desktop environment could vary. The real solution is probably to use `graphical-session.target`, but that only works when the desktop environments start using systemd to bring it up, which is a work in progress as it seems.

kakra:
[...]

PR #109, opened 2019-03-08
GPU performance level/mode pinning

mdiluz:
Applies on top of #108.

* Allows controlling `GPUPowerMizerMode` on Nvidia (`nv_powermizer_mode` in config)
* Allows controlling the `power_dpm_force_performance_level` file on AMD (`amd_performance_level` in config)
* Removes non-functioning `amd_core_clock_percentage` and `amd_mem_clock_percentage`, fully featured amd overclock tweaks can come later
* Implements properly getting the GPU state on AMD and Nvidia, so that GameMode always tries to return the state back to the way it found it

### DIFF
--- a/daemon/daemon_config.c
+++ b/daemon/daemon_config.c
@@ -92,8 +92,7 @@ struct GameModeConfig {
 		long nv_core_clock_mhz_offset;
 		long nv_mem_clock_mhz_offset;
 		long nv_perf_level;
-		long amd_core_clock_percentage;
-		long amd_mem_clock_percentage;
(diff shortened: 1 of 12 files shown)


OlliC:
Does this solve #45 ?

mdiluz:
I think that would have been this: https://github.com/FeralInteractive/gamemode/commit/949bf291bed40bc93ad50ade33c778c5fe8acf8e

aejsmith:
I've merged this manually since it includes the commits from #108.

mdiluz:
Cheers :)

PR #111, opened 2019-03-10
Expand on Nvidia config notes 

mdiluz:
`nv_powermizer_mode` needed a reference for the values, an `nv_perf_level` needed a little more explanation.


mdiluz:
Closing in favor of #112

PR #112, opened 2019-03-10
Clean up NV optimization config

mdiluz:
Expanding on #111 which will now be closed.

This adds more explanation about the nv config options, while also removing the need to specify the `nv_perf_level` attribute, as it was rightly seen to be a little confusing.

This also fixes some instances where the nv clock options weren't set.

### DIFF
--- a/daemon/daemon_config.c
+++ b/daemon/daemon_config.c
@@ -91,7 +91,6 @@ struct GameModeConfig {
 		long gpu_device;
 		long nv_core_clock_mhz_offset;
 		long nv_mem_clock_mhz_offset;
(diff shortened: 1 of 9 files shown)


aejsmith:
LGTM, thanks for cleaning that up. Do you want @rezzafr33 to test the last commit for #113 before I merge it, or is that OK to merge as is?

mdiluz:
You can merge, to my knowledge it's correct, but I'll follow up with a fix if it's not!

mdiluz:
Ha, got a response there at the moment, looks like the fix needs a little bit more - I'll follow up!

PR #121, opened 2019-03-19
Build with optimizations

lilianmoraru:
Note that `meson` builds [without optimizations by default](https://mesonbuild.com/Running-Meson.html#configuring-the-build-directory).
This turns `-g` into -> `-g -O2`(still keeps debug info, in case of a crash).


aejsmith:
Good call, thanks!

PR #122, opened 2019-03-27
Add more compiler warnings and fix the outfall, most importantly ensure strncpy are null terminated

gicmo:
Add some additional compiler warnings to catch more potential issues. `-Wstringop-truncation` lead to a few new warnings, some false positives, but I believe not all. I included some changes to make absolutely sure and obvious that all strings are properly null terminated. `'-Wshadow'` complained about `bus` being shadowed in `game_mode_inhibit_screensaver` and I spotted that the local `bus` was not unref'ed. `-Wformat-signedness` lead to a fixes for a few format strings. I have some cleanup patches for `inih` as well.

### DIFF
--- a/daemon/dbus_messaging.c
+++ b/daemon/dbus_messaging.c
@@ -269,21 +269,21 @@ int game_mode_inhibit_screensaver(bool inhibit)
 	const char *function = inhibit ? "Inhibit" : "UnInhibit";
 
 	sd_bus_message *msg = NULL;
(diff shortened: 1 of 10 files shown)


aejsmith:
Thanks! I've manually merged this and fixed up the clang-format complaints.

gicmo:
Indeed. I opend #123 for that, seems the compiler is a bit confused here.

PR #123, opened 2019-03-28
Rewrite strncpy to silence compiler warning

gicmo:
Instead of initializing the char array to 0 and writing N-1 bytes,
write N-1 bytes and set the last byte manually to 0. This seems
to make gcc happy.


aejsmith:
Thanks - I've done a 1.3.1 release which was needed to fix a compatibility issue with older Meson versions, and I've included both of your PRs in that as well.

PR #125, opened 2019-03-29
Use path_bindir for installing gamemoderun

tgurr:
Honor prefix for installing gamemoderun to properly work in cross
environment so it can be installed in e.g. /usr/x86_64-pc-linux-gnu/bin
instead of /usr/bin like it's already done for gamemoded.


aejsmith:
Merged, thanks!

PR #128, opened 2019-04-04
Gpuclockctl fixes

mgerstner:
I got a [review request](https://bugzilla.suse.com/show_bug.cgi?id=1129967) for an update of gamemode on openSUSE. The code used in the gpuclockctl that can be run as root via `pkexec` has a couple of issues that I want to address with these commits.


aejsmith:
All look good to me. Thanks!

PR #129, opened 2019-04-10
Introduce a ClientCount dbus Property

gicmo:
I want to write a simple GNOME shell extension that displays an icon if gamemode is active (maybe in the future a list of games requestion game mode as a menu would be nice too). For this a property that can be watched would be neat.

### DIFF
--- a/daemon/dbus_messaging.c
+++ b/daemon/dbus_messaging.c
@@ -182,22 +182,48 @@ static int method_query_status_by_pid(sd_bus_message *m, void *userdata,
 
 	return sd_bus_reply_method_return(m, "i", status);
 }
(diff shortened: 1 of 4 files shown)


aejsmith:
A shell extension is a good idea, let me know when you have that implemented. These changes look good to me, merged.

gicmo:
I have started with the [extension](https://github.com/gicmo/gamemode-extension). 

For now it will display an icon in that top bar, that will turn red once it is activated: gamemode-on icon

mdiluz:
Love it! This has always been a future TODO for me, glad someone has grabbed it much earlier than I would have :)

PR #136, opened 2019-05-02
Use readlink instead of realpath so find_exe works for flatpaks

gicmo:
Currently to find the executable for a game, a call to realpath(3) of /proc/<pid>/exe is made. The realpath(3) call will fail if the target does not exist (internally realpath will stat all the components of the link target path). This is a problem in the case of sandbox applications where the exe points to the absolute path *inside* the sandbox, e.g. to /app/bin/<name> in the case of flatpak. For these cases realpath(3) will then fail. Therefore use readlink(3) instead.

I have omitted the call to stat for the target because I am not sure it is necessary (and most likely the target will exist, because it just requested gamemode). I can add it though. Also by using readlink(3) we lose the ability to resolve names with `..` components but for our use case that should not matter.

### DIFF
--- a/daemon/gamemode.c
+++ b/daemon/gamemode.c
@@ -40,10 +40,14 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "helpers.h"
 #include "logging.h"
 
+#include <fcntl.h>
 #include <pthread.h>
 #include <signal.h>
 #include <stdatomic.h>
+#include <sys/stat.h>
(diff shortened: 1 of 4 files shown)


gicmo:
This is part of the ongoing efforts to make GameMode work with flatpak (see https://github.com/flathub/com.valvesoftware.Steam/issues/77).

aejsmith:
LGTM, thanks!

Issue #141, opened 2019-05-10
ioprio and renice options are not reset on unregister


mdiluz:
**Describe the bug**
In `game_mode_context_register` we set up these three optimisations:

```
	/* Apply scheduler policies */
	game_mode_apply_renice(self, client);
	game_mode_apply_scheduling(self, client);

	/* Apply io priorities */
	game_mode_apply_ioprio(self, client);
```

But there is no equivalent set of calls in `game_mode_context_unregister`. 

This will need a bit of nuance as the process may be changing these values itself, but all three should be able to be reset when appropriate. 

**To Reproduce**
Steps used to reproduce the behavior:
1. Set `ioprio` and `renice` in `gamemode.ini`
2. Request gamemode start from process
3. Request gamemode end from process
4. Note that I/O priority, niceness params are still set.

**Expected behavior**
Values should be re-set as appropriate.

mdiluz:
I've moved the `softrealtime` issue out to #139

mdiluz:
#142 has the fix for this

aejsmith:
Merged #142.

Issue #140, opened 2019-05-10
ioprio and renice options do not apply to full process tree of client


mdiluz:
**Describe the bug**
This is similar to https://github.com/FeralInteractive/gamemode/issues/139

When `ioprio` is set in `gamemode.ini`, we try to use `SYS_ioprio_set` here: https://github.com/FeralInteractive/gamemode/blob/master/daemon/gamemode-ioprio.c#L86

But, based on the man page for `ioprio_set` here: http://man7.org/linux/man-pages/man2/ioprio_set.2.html

>       Two or more processes or threads can share an I/O context.  This will
>       be the case when clone(2) was called with the CLONE_IO flag.
>       However, by default, the distinct threads of a process will not share
>       the same I/O context.  This means that if you want to change the I/O
>       priority of all threads in a process, you may need to call
>       ioprio_set() on each of the threads.  The thread ID that you would
>       need for this operation is the one that is returned by gettid(2) or
>       clone(2).

This implies a default thread creation won't inherit the `ioprio` value that has been set.

Interestingly the `renice` setting sets for the whole process tree, and as noted in the above man page:

>        If no I/O scheduler has been set for a thread, then by default the
>       I/O priority will follow the CPU nice value (setpriority(2)).

Which means this issue would be masked somewhat for those using `renice`.

**To Reproduce**
Steps used to reproduce the behavior:
1. Launch multiprocess program and request gamemode
2. Check ioprio value for each thread
3. See ioprio value has not been inherited by process tree

**Expected behavior**
The `ioprio` value should be inherited for the whole tree

**System Info (please complete the following information):**
 - PopOS 19.04
 - GameMode Version 1.3.1

mdiluz:
I've implemented tests in https://github.com/mdiluz/gamemode/commit/c082647416a98359ae0098d8a9f782e73b1b0180 to prove this issue.

mdiluz:
#142 has the fix for this

aejsmith:
Merged #142.

PR #142, opened 2019-05-11
ioprio and niceness fixes, and long cmdline options

mdiluz:
This PR has the following changes:

- ioprio and niceness fixes for issue #141 and #140
- support for long command line options, and requesting with `--request=PID` for a specific process

Help now looks like:
```
Usage: gamemoded [-d] [-l] [-r] [-t] [-h] [-v]

  -r[PID], --request=[PID] Toggle gamemode for process
                           When no PID given, requests gamemode and pauses
  -s[PID], --status=[PID]  Query the status of gamemode for process
                           When no PID given, queries the status globally
  -d, --daemonize          Daemonize self after launch
  -l, --log-to-syslog      Log to syslog
  -r, --test               Run tests
  -h, --help               Print this help
  -v, --version            Print version

See man page for more information.
```

I've also implemented tests for the two fixes above, both a302168 and a6e238d can be run to show the tests failing, before the subsequent fixes

### DIFF
--- a/daemon/daemon_config.c
+++ b/daemon/daemon_config.c
@@ -31,6 +31,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #define _GNU_SOURCE
 
 #include "daemon_config.h"
+#include "helpers.h"
 #include "logging.h"
 
 /* Ben Hoyt's inih library */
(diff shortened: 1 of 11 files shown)


mdiluz:
Edited title to reflect WIP - at the moment this doesn't cope with multiple clients, will need to sort that out.

mdiluz:
20d014b fixes the problem for now.

We don't mess with niceness or ioprio values that are non-default anyway, so there's no need to store the initial value.

A future change might be to make these values an offset, or something else, but I worry that's being too clever - if a program is adjusting them itself, we should probably not mess with it.

mdiluz:
I've now implemented multithreaded program fixes for ioprio and niceness optimizations, as well as tests for those implementations

PR #144, opened 2019-05-19
Add config hot-reloading

mdiluz:
This PR sets up two new ways to refresh the current configuration. For safety, if GameMode is active it tears down the current optimizations on both system and clients, swaps out to a new config, and then rebuilds the optimizations.

The changes provide two ways to do this:

- By watching the config locations with `inotify` for file changes, new config files, or moved and deleted config files. Currently uses the reaper thread to check the inotify file descriptor for updates.
- By using the new `RefreshConfig` dbus interface (not exposed in `gamemode_client.h` for now).

The dbus interface is implemented as a fallback, although perhaps an argument could be made that "supervisor" clients should be allowed to call this if ever needed, and a command line argument to `gamemoded` could be implemented alongside that.

### DIFF
--- a/daemon/dbus_messaging.c
+++ b/daemon/dbus_messaging.c
@@ -213,7 +213,7 @@ static int method_refresh_config(sd_bus_message *m, void *userdata,
                                  __attribute__((unused)) sd_bus_error *ret_error)
 {
 	GameModeContext *context = userdata;
-	int status = game_mode_refresh_config(context);
+	int status = game_mode_reload_config(context);
 	return sd_bus_reply_method_return(m, "i", status);
 }
 
--- a/daemon/dbus_messaging.c
+++ b/daemon/dbus_messaging.c
@@ -220,22 +220,21 @@ static int method_refresh_config(sd_bus_message *m, void *userdata,
 /**
  * D-BUS vtable to dispatch virtual methods
  */
-static const sd_bus_vtable gamemode_vtable[] = {
-	SD_BUS_VTABLE_START(0),
-	SD_BUS_PROPERTY("ClientCount", "i", property_get_client_count, 0,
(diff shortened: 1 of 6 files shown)


aejsmith:
When reloading the config, the reaper thread will disable optimisations then re-enable them while holding a write lock. However, `game_mode_context_(un)register` do not hold the lock around enabling/disabling optimisations. Seems like there's a race condition there if we get a client (un)register at the same time the reaper reloads the config?

mdiluz:
Good point - I'd assumed that to be the case, but forgot that it was only the dbus single-threaded message queue that was protecting that race condition before. 4612533 should resolve that and passes the tests.

daemon/dbus_messaging.c
```diff
@@ -206,24 +206,35 @@ void game_mode_client_count_changed(void)
...
-	SD_BUS_METHOD("QueryStatus", "i", "i", method_query_status, SD_BUS_VTABLE_UNPRIVILEGED),
-	SD_BUS_METHOD("RegisterGameByPID", "ii", "i", method_register_game_by_pid,
-	              SD_BUS_VTABLE_UNPRIVILEGED),
-	SD_BUS_METHOD("UnregisterGameByPID", "ii", "i", method_unregister_game_by_pid,
-	              SD_BUS_VTABLE_UNPRIVILEGED),
-	SD_BUS_METHOD("QueryStatusByPID", "ii", "i", method_query_status_by_pid,
-	              SD_BUS_VTABLE_UNPRIVILEGED),
-	SD_BUS_VTABLE_END
-};
+static const sd_bus_vtable gamemode_vtable[] =
```
mdiluz: Note this keeps getting changed by slightly differing `clang-format` versions. Perhaps we can do something about that eventually...

Issue #51, opened 2018-05-12
Missing a sysv init script


NotMrFlibble:
Not all of us use systemd…

chewi:
It's not that simple, the service itself actually requires systemd though it could be made to work without it. See #15.

edddeduck:
@NotMrFlibble As per @mdiluz comment (see https://github.com/FeralInteractive/gamemode/issues/15) the design has a clear separation, so it's entirely possible for someone to come and program a non-systemd messaging backend that'll still work with all the same client code. 

If you or someone else write one we'll happily test and merge a PR that has this functionality but given the resources available and the smaller number of people requesting it we don't have any plans to write one in the immediate future.

esmithferal:
@NotMrFlibble As per @mdiluz comment (see https://github.com/FeralInteractive/gamemode/issues/15) the design has a clear separation, so it's entirely possible for someone to come and program a non-systemd messaging backend that'll still work with all the same client code.

If you or someone else write one we'll happily test and merge a PR that has this functionality but given the resources available and the smaller number of people requesting it we don't have any plans to write one in the immediate future.

Issue #15, opened 2018-04-10
Build with with-systemd=false fails


kozec:
Hello,

I've found that building with `with-systemd` build flag disabled fails, as it still declares dependency on libsystemd library:

```
$ meson --prefix=/usr build -Dwith-systemd=false
The Meson build system
Version: 0.45.1
Source dir: /tmp/gamemode/src/gamemode-1.0
Build dir: /tmp/gamemode/src/gamemode-1.0/build
Build type: native build
Project name: gamemode
Native C compiler: cc (gcc 7.3.1 "cc (GCC) 7.3.1 20180312")
Build machine cpu family: x86_64
Build machine cpu: x86_64
Found pkg-config: /usr/bin/pkg-config (0.29.2)

meson.build:39:0: ERROR: Native dependency 'libsystemd' not found
```

mdiluz:
Thanks @kozec,

Does this patch resolve the issue for you?

```
diff --git a/meson.build b/meson.build
index 1680d7e..b937881 100644
--- a/meson.build
+++ b/meson.build
@@ -36,7 +36,7 @@ path_libdir = join_paths(path_prefix, get_option('libdir'))
 path_libexecdir = join_paths(path_prefix, get_option('libexecdir'))
 
 # Find systemd via pkgconfig
-dep_systemd = dependency('libsystemd')
+dep_systemd = dependency('libsystemd', required: false)
 
 # Allow meson to figure out how the compiler sets up threading
 dep_threads = dependency('threads')
```

eNTi:
hi. i'm trying to get this to work without systemd as well:
```
==> Starting build()...
+ exec meson setup --prefix /usr --libdir /usr/lib --libexecdir /usr/lib --bindir /usr/bin --sbindir /usr/bin --includedir /usr/include --datadir /usr/share --mandir /usr/share/man --infodir /usr/share/info --localedir /usr/share/locale --sysconfdir /etc --localstatedir /var --sharedstatedir /var/lib --buildtype release --wrap-mode nofallback gamemode-1.0 build -Dwith-systemd=false -D b_lto=true
The Meson build system
Version: 0.45.1
Source dir: /tmp/makepkg/gamemode/src/gamemode-1.0
Build dir: /tmp/makepkg/gamemode/src/build
Build type: native build
Project name: gamemode
Native C compiler: cc (gcc 7.3.1 "cc (GCC) 7.3.1 20180312")
[...]

PR #147, opened 2019-05-24
Ability to use libdbus for client's dbus messaging

gicmo:
In flatpaks systemd is not available, and thus would need to be build as an extra dependency (see [com.valvesoftware.Steam#180](https://github.com/flathub/com.valvesoftware.Steam/pull/180/files)). For the [GameMode Tester](https://github.com/gicmo/gamemode-tester) App, I re-implemented the client bits and extracted those into a separate project: [gamemode-libdbus](https://github.com/gicmo/gamemode-libdbus). The code from this PR is basically taking from there and adds support for using `libdbus`.
I am a bit unsure what is the best way to go:
   - Add `libdbus` support to GameMode (this PR):
      - pro: central place where code is located
      - con: having  `#ifdef`/`if ! systemd_dep.found()` sprinkled all over the place and two basically different client libs in one code-base
  - Use  [gamemode-libdbus](https://github.com/gicmo/gamemode-libdbus) in flatpaks
      - pro: small dependency for flatpaks
      - con: having to maintain a separate client implementation (maybe not that bad?)
  - Always use `libdbus` in GameMode instead of systemd
      - pro: no `ifdef`/`if`-hell
      - pro: central location of all client code
      - con: tow different dbus APIs present in one project (maybe not so bad?)

What do you guys think?

### DIFF
--- a/lib/client_impl.c
+++ b/lib/client_impl.c
@@ -31,6 +31,7 @@ POSSIBILITY OF SUCH DAMAGE.
 
 #define _GNU_SOURCE
 
+#include <dbus/dbus.h>
 #include <dlfcn.h>
 #include <errno.h>
 #include <stdio.h>
@@ -38,8 +39,10 @@ POSSIBILITY OF SUCH DAMAGE.
 #include <string.h>
 #include <sys/stat.h>
 #include <sys/types.h>
-#include <systemd/sd-bus.h>
 #include <unistd.h>
+
+// For developmental purposes
+#define DO_TRACE 0
 
 // D-Bus name, path, iface
 #define DAEMON_DBUS_NAME "com.feralinteractive.GameMode"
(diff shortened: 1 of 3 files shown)


mdiluz:
So I suppose I should explain that I used systemd because I felt its api was far more elegant and much easier to use. I also could actually get the damn thing to work, where libdbus seemed to constantly give pain.

Given libdbus-1 and systemd are effectively interchangeable (correct?) I think swapping the client lib to libdbus seems fine. We'll have two dbus implementations at once but oh well.

If you wanted to PR a swap the daemon to dbus as well wouldn't see that as a bad change either as long as it was cleanly written and thoroughly tested. Just maybe wait a bit - I have one hell of a project cleanup/reorganisation PR incoming...

gicmo:
@mdiluz very cool. I will change this PR to switch the client and then sit tight until your cleanup work is landed!

mdiluz:
Also side note: GameMode Tester is very cool. Was wondering if I needed to make the testing code a bit less of a hack because currently it only allows testing if there isn't a portal between the two. This at least helps a ton, but I'll maybe do a PR sometime to refactor testing into something more broadly usable.

cwilling:
Does this change help other more general non-systemd use cases (not necessarily just flatpack usage)? I see #15 and #51 have already asked about this - perhaps this change is a possible solution. If so, very useful.

mdiluz:
@cwilling if the daemon is swapped over too, then yes. Though I think there'll be a little nuance in making sure it's a positive solution for both user and developer experience.

aejsmith:
I'm happy with doing this for the client library now, looks good to me and I can't see any regressions so I've merged it. Noticed just _after_ merging that you'd marked it as draft though... did you have more planned changes?

gicmo:
@aejsmith Sorry for the late answer. Nope, was good to go. Just forgot to remove the draft status. Thanks for merging! I will convert the server to libdbus too, once #148 is merged.

cwilling:
Great @aejsmith, looking forward to it. I was going to try it myself although without any previous experience of dbus API. It's definitely better to be done by someone who knows what they're doing :)

PR #148, opened 2019-05-25
Spring cleaning

mdiluz:
- README restructure and cleanup
- Source file rename and restructure
- Wine code full separation into wine file
- Include cleanup
- Comment cleanup


common/common-gpu.h
```diff
@@ -30,7 +30,7 @@ POSSIBILITY OF SUCH DAMAGE.
  */
 
 #pragma once
-#include "daemon_config.h"
+#define CONFIG_VALUE_MAX 256
```
aejsmith: Why the duplicated definition?

daemon/gamemode-context.c
```diff
@@ -602,10 +602,12 @@ int game_mode_context_query_status(GameModeContext *self, pid_t client, pid_t re
  */
 static GameModeClient *game_mode_client_new(pid_t pid, char *executable)
 {
+	/* clang-format off */
```
aejsmith: Why is this needed (and the other one)? These are in my preferred style already but is it not possible to get clang-format to do this?

aejsmith:
Added a couple of comments, and can you rebase on master since there are now conflicts?

daemon/gamemode-context.c
mdiluz: I'm not 100% sure why but on multiple different installs, clang reformats these two sections differently to the Travis instance the CI is running. I've tried to coax it into doing it correctly with options but nothing seems to work.

common/common-gpu.h
mdiluz: To split out the dependency between `common` and `daemon` code.

But yeah, it's a bit confusing now, I'll clean it up.

mdiluz: Fixed in d9b9141

mdiluz:
Rebased and addressed comments

PR #151, opened 2019-06-01
Turn off debug logging in client impl

mdiluz:
Accidentally left in #147


aejsmith:
Oops, didn't notice that and thought the debug logging was intentional on a debug build.

I've also followed that up with c64aa5e5b76a745d92aa2748ead629908e1ae580, which always enables the error logging rather than only when DO_TRACE is enabled.

mdiluz:
Ha! Dang didn't notice that part as well. Good Spot

PR #152, opened 2019-06-02
added option to disable building the utils

ysblokje:
When you need both 32 and 64 bit version of the library only one version of the utils is required. I've created an option to disable building the util folder. 
This option is with-util and is by default set to true, so it won't break any normal builds.


mdiluz:
Ah cheers, good point

aejsmith:
Merged, thanks!

PR #155, opened 2019-07-02
Expose registered games on the bus

gicmo:
For the Shell extension (or a little separate tool) as well as future integration with [GNOME Usage](https://gitlab.gnome.org/GNOME/gnome-usage) I wanted to get the list of currently registered games from gamemoded. For this I added to the `com.feralinteractive.GameMode` interface:
```
methods:
  ListGames: Array of Tuples (ProcessId, ObjectPath)

signals:
  GameRegistered: ProcessId, ObjectPath
  GameUnregistered: ProcessId, ObjectPath
```
As well as a new `com.feralinteractive.GameMode.Game` interface:
```
properties:
  ProcessId: int
  Executable: string
```
This should make it possible for a client to track all currently registered games (start listening to the signals, list the current active ones and track changes via the signals). 

In order to expose the individual games, I opted for expose the `GameModeClient` struct (as an opaque struct with getters). To be able to do this I added reference counting to it, so it can outlive being reaped by the reaper. Since the object is de-facto immutable this should be sufficient.

### DIFF
--- a/daemon/gamemode-context.c
+++ b/daemon/gamemode-context.c
@@ -87,7 +87,6 @@ static GameModeContext instance = { 0 };
 static volatile bool had_context_init = false;
 
 static GameModeClient *game_mode_client_new(pid_t pid, char *exe);
-static void game_mode_client_unref(GameModeClient *client);
 static const GameModeClient *game_mode_context_has_client(GameModeContext *self, pid_t client);
 static void *game_mode_context_reaper(void *userdata);
 static void game_mode_context_enter(GameModeContext *self);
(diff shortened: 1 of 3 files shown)


gicmo:
Just realized I have some inconsistencies between pids on the bus as unsigned `u` vs pids as signed `i`. Will fix.

aejsmith:
All looks good to me, merged. Thanks!

I did one follow up change (12cf304ff2c89e1d0bb9d310344e6a61bc0dc929) just to reduce some duplication of the game path prefix string.

gicmo:
Thanks! The follow-up change makes total sense. Btw, any plans for a 1.4 release? ;)

aejsmith:
Don't have any specific plans for a new release currently, but since there's a reasonable amount of new stuff I'll see if we can do one soon.

PR #158, opened 2019-07-12
daemon: properly report error in error case

gicmo:
Currently, in method_list_games (`ListGames` on the bus), when `sd_bus_message_append` fails, we break the loop but then we were not reporting the error (contained in `r`) back to call.


gicmo:
I just realized there was also a spelling mistake in the `Executable` property. Ups. Sorry about that.

aejsmith:
Well, somehow I managed to not notice the typo either despite looking at QDBusViewer to confirm the new properties were working, so no problem :)

Merged both fixes, thanks.

PR #159, opened 2019-07-12
Add hybrid GPU support

### LINKED ISSUE #20: Support for hybrid graphic cards
Bumblebee, nvidia-prime and all the other options that exist

stephanlachnit:
I added hybrid GPU support by adding ```$GAMEMODERUNEXEC``` to the gamemoderun launch command. If ```GAMEMODERUNEXEC``` is unset, which given by its very specific name always should be the case by default, the program will execute normal. However if you set the environment variable to a specific command like ```optirun```, it will start optirun, which starts the game.
Since the script injects libgamemodeauto globally inside the script, it will inject GameMode to the wrapper as well as the game. If a wrapper is not important for the performance (like ```env XY=Z```), the user can blacklist it.

closes #20 closes #87 

Signed-off-by: Stephan Lachnit <stephanlachnit@protonmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,14 @@ Config files are loaded and merged from the following directories, in order:
 4. `$PWD`
 
 ---
+## Note for Hybrid GPU users
+
+It's not possible to integrate commands like optirun automatically inside GameMode, since GameMode only starts if the game is already started.
+However it is possible to use a hybrid GPU wrapper like optirun by starting the game with gamemoderun.
+You can do this by setting the environment variable `GAMEMODERUNEXEC` to your wrapper's launch command, so for example `GAMEMODERUNEXEC=optirun` or `GAMEMODERUN="env DRI_PRIME=1"`.
+GameMode will not be injected to the wrapper.
+
+---
 ## Apps with GameMode integration
 
 ### Games
--- a/data/gamemoded.8.in
+++ b/data/gamemoded.8.in
@@ -56,6 +56,12 @@ The library can be manually preloaded if needed:
 LD_PRELOAD=$LD_PRELOAD:/usr/\e$LIB/libgamemodeauto.so.0 ./game
 .RE
 
+It is possible to set additional start commands to gamemoderun by setting the environment variable:
+
+.RS 4
+GAMEMODERUNEXEC="command"
+.RE
+
 The \fBgamemode_client.h\fR header can be used by developers to build the requests into a program:
 
 .RS 4
--- a/data/gamemoderun.in
+++ b/data/gamemoderun.in
@@ -5,6 +5,6 @@
 # appropriate path depending on whether the app is 32- or 64-bit.
 GAMEMODEAUTO="@GAMEMODE_PREFIX@/\$LIB/libgamemodeauto.so.0"
 
-export LD_PRELOAD="${GAMEMODEAUTO}${LD_PRELOAD:+:$LD_PRELOAD}"
+LD_PRELOAD="${GAMEMODEAUTO}${LD_PRELOAD:+:$LD_PRELOAD}"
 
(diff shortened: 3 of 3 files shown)


aejsmith:
Couldn't you just prefix your game launch commands with whichever wrapper you need as well as gamemoderun, e.g. `optirun gamemoderun %command%`? That way you wouldn't even need to blacklist the wrapper since that wouldn't get `LD_PRELOAD` set for it.

I suppose the benefit of doing it this way though is that you can set the environment variable globally and not have to duplicate it through all of your Steam game launch options or anything like that.

stephanlachnit:
> Couldn't you just prefix your game launch commands with whichever wrapper you need as well as gamemoderun, e.g. `optirun gamemoderun %command%`? That way you wouldn't even need to blacklist the wrapper since that wouldn't get `LD_PRELOAD` set for it.
> 
> I suppose the benefit of doing it this way though is that you can set the environment variable globally and not have to duplicate it through all of your Steam game launch options or anything like that.

The main reason is the global environment variable. I know this is a bit optimistic, but imagine programs like Lutris or Steam would add support for gamemoderun, so that you could simply toggle it in the settings and don't have to edit every start command.

And even if this will never happen, writing a command one time to the blacklist and add one global environment variable is shorter than writing it to every launch command.
After all, gamemoderun is there to make life easier.

And some wrappers can impact the performance. For example primus does transport frames from a virtual secondary display to the actual output display. If you focus all performance on the game, this could potentially cause stuttering (although I didn't test this hypothesis).

Overall I think while any hybrid gpu solution (especially with nvidia) is never perfect, at the moment where we can't choose the GPU for rendering ingame, this is a simple solution.

mdiluz:
I like the simplicity here - rather than going through the config or anything else, simply setting an env variable. It's not perfect, but I think the fact it's a "dumb" feature, not trying to do anything too clever, protects it from most issues I can come up with.

One note: I wonder if the wrapper backlist could be worked around by using something like this?
```
_LD_PRELOAD="${CONFIG_LIB_DIR}${LD_PRELOAD:+:$LD_PRELOAD}"

exec $GAMEMODERUNEXEC env LD_PRELOAD=${_LD_PRELOAD} "$@"
```

stephanlachnit:
> One note: I wonder if the wrapper backlist could be worked around by using something like this?

I also thought about that. The problem is that now the wrapper will run ```env``` (I checked it with the debug option from optirun).
The program will still be run on the secondary gpu, at least from my testing with glxinfo, and my very basic benchmarking with glxgears implies that there isn't a performance impact.

I don't think one way is better than the other. If you managed to set up your hybrid gpu on Linux in the first place, it shouldn't be an obstacle to set your wrapper to the blacklist. I personally wouldn't do it with my current configuration of optirun, but I wouldn't mind to change the code if you think that this a cleaner solution.

mdiluz:
`env` is a command line utility that won't use the GPU at all, so I don't think any of primus and friends will have any impact on it at all

stephanlachnit:
> `env` is a command line utility that won't use the GPU at all, so I don't think any of primus and friends will have any impact on it at all

I thought more about the way optirun etc works. If they just activate the gpu and set some environment variables it doesn't make any difference, which I suspect is the way how they work.
However I could also imagine that some wrappers want to now the pid or the executable for some reason. In this case they don't get the right one but instead the one from the env parent process.

As I said, I didn't see a difference with my testing. Since most people either use PRIME, primus or optirun, which all work basically with environment variables, it doesn't really matter. Just tell me if I should change it.

aejsmith:
Thanks, I've merged this and followed it up with 50303dc09299f8e7ec015a85b7048fecd47e9796 to expand the documentation a bit.

PR #161, opened 2019-07-14
Add and expose Timestamp and Requester properties for GameModeClient

gicmo:
Record the timestamp a client was created and who requested it and expose them as `Timestamp` and `Requester` properties of the `com.feralinteractive.GameMode.Game` interface. I am thinking of creating a small `gamemodeclt` command line tool to show the current status, and list all the clients and all the information we got about them. Maybe also a `gamemodectl -p <pid>` to request GameMode on behalf of `pid`. What do you guys think?

### DIFF
--- a/daemon/gamemode-context.c
+++ b/daemon/gamemode-context.c
@@ -46,6 +46,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #include <pthread.h>
 #include <stdatomic.h>
 #include <stdlib.h>
+#include <sys/time.h>
 #include <systemd/sd-daemon.h> /* TODO: Move usage to gamemode-dbus.c */
 #include <unistd.h>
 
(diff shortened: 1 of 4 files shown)


daemon/gamemode.h
```diff
@@ -69,6 +70,11 @@ pid_t game_mode_client_get_pid(GameModeClient *client);
  */
 const char *game_mode_client_get_executable(GameModeClient *client);
 
+/**
+ * The path to the executable of client.
```
aejsmith: Wrong comment.

daemon/gamemode-context.c
```diff
@@ -714,6 +726,15 @@ const char *game_mode_client_get_executable(GameModeClient *client)
 	return client->executable;
 }
 
+/**
+ * The path to the executable of client.
```
aejsmith: Wrong comment.

aejsmith:
Just a couple of incorrect function comments, otherwise looks good to me. I'll fix these up and merge.

As for `gamemodectl`, I like the idea. Note that you can already use `gamemoded` to request game mode, either by making the request from that process and waiting until interrupted, or on behalf of another process. However, if a separate `gamemodectl` tool is added, I think that'd be a better place for that functionality to live.

PR #165, opened 2019-07-29
Add Lutris instructions

Aurnytoraink:
It works with native games, windows games (using wine), and tested with Wii games (using dolphin emulator)

PS: It doesn't work on Steam games (no break, the gamemode is just not activate)


Aurnytoraink:
Anything new ?
Do the mainteners are in vacation ? 😂

aejsmith:
Sorry for the delay, merged. Thanks!

PR #166, opened 2019-07-30
Added: Instruction on how to enable gamemode for all Proton games

toluschr:
This is just what I use! I don't know if this is useful, but I still want to share it with you.

The simple python script is important for the steam overlay to work, since it gets loaded via `LD_PRELOAD` too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,19 @@ Or edit the Steam launch options:
 ```bash
 gamemoderun %command%
 ```
+If you want gamemode to be active for **all** Proton games, you can set the `user_settings.py` (Usually in `steamapps/common/Proton/`) to the following:
+```
+import os
+
+user_settings = {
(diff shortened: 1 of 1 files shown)


mdiluz:
Locally this only appears to work for a small period of time during launch, I'm not completely sure why. Tested with a couple of different games.

If it does work and can be refined, then it might need a note that it needs to be added for every individual version of Proton. Perhaps there's a way to enact that globally?

kakra:
This would be fixed by #76 because you could just run the full Steam client with `gamemoderun`. However, I see that my branch needs rework because we also have integration with other launchers now (tho, those usually have out-of-the-box support for GameMode).

kakra:
BTW: Should this be closes? The original repository to pull from even doesn't exist any longer.

PR #173, opened 2019-10-07
Introduce a new pidfd based set of D-Bus APIs

gicmo:
Pidfds are process file descriptors representing processes. Since Linux [5.2](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit?id=b3e5838252665ee4cfa76b82bdf1198dca81e5be) the corresponding pid in the callers pid namespace can be obtained via the fdinfo in procfs; this by reading the "Pid:" field in '/proc/self/fdinfo/<pidfd>'. In Linux 5.3 the [`pidfd_open`](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit?id=32fcb426ec001cb6d5a4a195091a8486ea77e2df) system call was added, allowing us to easily open a pidfd for a given pid. Additionally select/poll/epoll can be used to monitor the status of the process, which can be used to watch for process termination.
Add a new set of APIs, analogous to the existing *byPID APIs, that work with pidfds instead of pids and convert the client library to default to using those, if supported by the system. The two main advantages are:

  - pidfds drastically simplify the container use case, because there is no need to translate the pids between different namespaces anymore

  - it prepares the use of pidfds in the daemon, which in the future could be used to monitor process terminations and thus removes the need of the reaper background thread.

### DIFF
--- a/daemon/gamemode-dbus.c
+++ b/daemon/gamemode-dbus.c
@@ -169,7 +169,7 @@ static int method_unregister_game_by_pid(sd_bus_message *m, void *userdata,
 }
 
 /**
- * Handles the QueryStatus D-BUS Method
+ * Handles the QueryStatusByPID D-BUS Method
  */
 static int method_query_status_by_pid(sd_bus_message *m, void *userdata,
                                       __attribute__((unused)) sd_bus_error *ret_error)
@@ -188,8 +188,9 @@ static int method_query_status_by_pid(sd_bus_message *m, void *userdata,
 
 	return sd_bus_reply_method_return(m, "i", status);
 }
+
 /**
- * Handles the Active D-BUS Method
+ * Handles the ClientCount D-BUS Property
  */
 static int property_get_client_count(sd_bus *local_bus, const char *path, const char *interface,
                                      const char *property, sd_bus_message *reply, void *userdata,
--- a/daemon/gamemode-dbus.c
+++ b/daemon/gamemode-dbus.c
@@ -32,7 +32,9 @@ POSSIBILITY OF SUCH DAMAGE.
 #define _GNU_SOURCE
 
 #include "gamemode.h"
+#include "common-helpers.h"
 #include "common-logging.h"
+#include "common-pidfds.h"
 
 #include <systemd/sd-bus.h>
 #include <systemd/sd-daemon.h>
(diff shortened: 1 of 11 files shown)


daemon/gamemode-dbus.c
```diff
@@ -189,6 +191,84 @@ static int method_query_status_by_pid(sd_bus_message *m, void *userdata,
 	return sd_bus_reply_method_return(m, "i", status);
 }
 
+/**
+ * Handles the RegisterGameByPID D-BUS Method
```
aejsmith: Incorrect method name on all of these comments (should be ByPIDFd).

daemon/gamemode-dbus.c
```diff
@@ -189,6 +191,84 @@ static int method_query_status_by_pid(sd_bus_message *m, void *userdata,
...
+	pid_t pids[2] = { 0, 0 };
+	GameModeContext *context = userdata;
+
+	int ret = sd_bus_message_read(m, "hh", &fds[0], &fds[1]);
+	if (ret < 0) {
+		LOG_ERROR("Failed to parse input parameters: %s\n", strerror(-ret));
+		return ret;
+	}
+
+	int reply = pidfds_to_pids(fds, pids, 2);
```
aejsmith: Are the FDs being leaked here? I don't see a close for them. Same for the other methods.

daemon/gamemode-dbus.c
gicmo: Ups. :+1:

daemon/gamemode-dbus.c
gicmo: I believe `sd_bus_message_read` won't `dup` the file descriptors so they are being closed when the message is freed. Let me verify this though. The mapage is silent on it :/

aejsmith:
Sorry for the delay. I added a few comments on one of the commits, could you check those?

Other than that, it seems to work for me on both kernel 5.2 (without pidfd_open) and 5.3.

daemon/gamemode-dbus.c
gicmo: Confirmed with @poettering, doesn't dup, we should be good.

aejsmith:
I just found that the supervisor tests (`gamemoded -t`) fail when using pidfds on kernel 5.3. I think it's requesting GameMode for the wrong PID - looks as though it gets applied for the PID that makes the request, not the PID it's requesting GameMode for. Could you have a look at that?

gicmo:
> I just found that the supervisor tests (`gamemoded -t`) fail when using pidfds on kernel 5.3. I think it's requesting GameMode for the wrong PID - looks as though it gets applied for the PID that makes the request, not the PID it's requesting GameMode for. Could you have a look at that?

I see it too, let me have a look

gicmo:
Ok, I had `callerpid` and `clientpid` swapped in the client library. Doh. Should be fixed now. Had to increase the wait timeout in one test for it to pass here (the child and the parent are racing for RegisterGame and QueryStatus).

aejsmith:
That's working for me now, merged. Thanks!

I was also seeing an occasional failure on "Gamemoderun and reaper thread tests", if that's what you were getting. I've increased the timeout for that.

gicmo:
> I was also seeing an occasional failure on "Gamemoderun and reaper thread tests", if that's what you were getting. I've increased the timeout for that.

Great. I did that too locally but did not push the commit apparently. Thanks for that and the review!

PR #198, opened 2020-02-24
Don't default to syslog logging

seb128:
Let the preferred/configured logging system to be used instead

### DIFF
--- a/data/gamemoded.service.in
+++ b/data/gamemoded.service.in
@@ -5,7 +5,7 @@ Description=gamemoded
 Type=dbus
 BusName=com.feralinteractive.GameMode
 NotifyAccess=main
-ExecStart=@BINDIR@/gamemoded -l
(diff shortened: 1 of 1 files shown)


seb128:
The issue was raising in the Ubuntu review for promotion to main
https://bugs.launchpad.net/ubuntu/+source/gamemode/+bug/1853830

`However, I don’t understand why the systemd service is called with -l (log to syslog), which is out default for systemd service (the journal forwards to syslog, even for user services). The day we want to remove syslog completely, it will prevent having to patch it.`

That's a valid point, forwaring to syslog is configured by the distribution/sysadmin when wanted, there is no reason individual services need to enforce the option

mdiluz:
Ah, my bad here - I'd not known the defaults would correctly forward the output if it wasn't specified

PR #199, opened 2020-02-26
prefer system installation of inih

### LINKED ISSUE #195: Remove inih subproject
Currently working on getting gamemode 1.5 packed for Debian, and we try to put our build files just on top of this repo, as it is good practice to keep track of all changes. However the inih submodule is really a struggle for external build systems, and also not really that nice considering how old the submodule state in this repo really is (2 years old). Since inih is packed for Debian and Fedora ([pkgs.org](https://pkgs.org/download/inih)), I don't see any good reason to keep this.
Could you do this? I could even write the PR if you don't have the time for it.

stephanlachnit:
With https://github.com/benhoyt/inih/pull/97 merged, it will be easy to install inih as a system library and declare the dependency with pkg-config. gamemode should use the system library instead of including the source, but we can still use it as a fallback ([see meson reference](https://mesonbuild.com/Subprojects.html#toggling-between-system-libraries-and-embedded-sources)).

Closes #195.

### DIFF
--- a/daemon/gamemode-config.c
+++ b/daemon/gamemode-config.c
@@ -36,7 +36,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "common-logging.h"
 
 /* Ben Hoyt's inih library */
-#include "ini.h"
+#include <ini.h>
 
 #include <dirent.h>
 #include <math.h>
--- a/meson.build
+++ b/meson.build
@@ -157,8 +157,10 @@ endif
 # main library
 if with_daemon == true
     # inih currently only needed by the daemon
-    inih = subproject('inih')
-    inih_dependency = inih.get_variable('inih_dependency')
+    inih_dependency = dependency(
+        'inih',
+        fallback : ['inih', 'inih_dependency']
+    )
 
     subdir('daemon')
 


stephanlachnit:
Btw it looks like Travis is broken? Build works fine, but "failes" because of
```
chmod: cannot access 'git-clang-format': No such file or directory
git-clang-format not downloaded
The command "./scripts/format-check.sh" exited with 1.
```

mdiluz:
Thanks for this, and explaining the needs in the issue.

It does seem like travis is having some fun again... [this line](https://github.com/FeralInteractive/gamemode/blob/master/scripts/format-check.sh#L7) should have downloaded it, so maybe it just needs a re-run... @afayaz-feral?

afayaz-feral:
A re-run worked fine, not entirely sure why travis decided to do that.

Thanks for this, happy to merge it.

mdiluz:
Tbh we don't check the wget for errors do we? Might as well just add the script the the repo or something

afayaz-feral:
I'm definitely not a fan of downloading a script that could change and break things without warning. I'll consider some alternatives.

PR #201, opened 2020-02-28
use upstream inih r48

stephanlachnit:
I added meson support to inih, so we can use upstream inih.
Only drawback is that gamemode will now install `libinih.so.47` if it is in fallback mode (i.e. linking to the shared library and not against the static), but thanks to the meson build file in upstream inih it shouldn't be hard to pack inih for distro maintainers.
 It's probably a good idea to update the submodule to inih r48 once it is released and leave it there (should be fallback only anyway).

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,7 +85,7 @@ GameMode depends on `meson` for building and `systemd` for internal communicatio
 
 #### Ubuntu/Debian (you may also need `dbus-user-session`)
 ```bash
-apt install meson libsystemd-dev pkg-config ninja-build git libdbus-1-dev
+apt install meson libsystemd-dev pkg-config ninja-build git libdbus-1-dev libinih-dev
 ```
 #### Arch
 ```bash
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -85,7 +85,7 @@ GameMode depends on `meson` for building and `systemd` for internal communicatio
 
 #### Ubuntu/Debian (you may also need `dbus-user-session`)
 ```bash
-apt install meson libsystemd-dev pkg-config ninja-build git libdbus-1-dev
+apt install meson libsystemd-dev pkg-config ninja-build git libdbus-1-dev libinih-dev
```
stephanlachnit: Btw right now this actually isn't a real dependency, `libinih-dev` doesn't have a pkg-config entry in Debian (yet).

stephanlachnit:
Maybe wait for https://github.com/benhoyt/inih/pull/100

stephanlachnit:
Updated against inih r48. I also removed `git submodule update` from the script, I think it should stay there. Anyway, in case you merge this and there are no other code updates planned, can you release this maybe as a `1.5.1`? Would allow us to do a Debian release with proper inih dependencies.

afayaz-feral:
Thanks, merged. I'll do a new 1.5.1 release for this soon.

Issue #204, opened 2020-03-08
Cannot install gamemode 1.5.1 on Linux Mint 19.3


lordairivis:
**Describe the bug**
Unable to install gamemode 1.5.1 on Linux Mint 19.3 (Ubuntu 18.04)

**To Reproduce**
1. Be running latest up-to-date Linux Mint 19.3
2. `git clone` gamemode from github
3. `cd gamemode`
4. `./bootstrap.sh`
5. See error

**Expected behavior**
Successful build

**System Info (please complete the following information):**
 - OS and version: Linux Mint 19.3
 - GameMode Version: 1.5.1

**Additional context**
```
~/gamemode $ apt-cache policy libinih-dev
libinih-dev:
  Installed: 40-1
  Candidate: 40-1
  Version table:
 *** 40-1 500
        500 http://la-mirrors.evowise.com/ubuntu bionic/universe amd64 Packages
        100 /var/lib/dpkg/status
```

```
~/gamemode $ apt-cache policy meson
meson:
  Installed: 0.45.1-2ubuntu0.18.04.2
  Candidate: 0.45.1-2ubuntu0.18.04.2
  Version table:
 *** 0.45.1-2ubuntu0.18.04.2 500
        500 http://la-mirrors.evowise.com/ubuntu bionic-updates/universe amd64 Packages
        500 http://la-mirrors.evowise.com/ubuntu bionic-updates/universe i386 Packages
        500 http://security.ubuntu.com/ubuntu bionic-security/universe amd64 Packages
        500 http://security.ubuntu.com/ubuntu bionic-security/universe i386 Packages
        100 /var/lib/dpkg/status
     0.45.1-2 500
        500 http://la-mirrors.evowise.com/ubuntu bionic/universe amd64 Packages
        500 http://la-mirrors.evowise.com/ubuntu bionic/universe i386 Packages
```
Full output of `./bootstrap.sh`:
```
~/gamemode $ ./bootstrap.sh 
+ meson --prefix=/usr build --buildtype debugoptimized -Dwith-systemd-user-unit-dir=/etc/systemd/user
The Meson build system
Version: 0.45.1
Source dir: /home/lord/gamemode
Build dir: /home/lord/gamemode/build
Build type: native build
Project name: gamemode
Native C compiler: cc (gcc 7.4.0 "cc (Ubuntu 7.4.0-1ubuntu1~18.04.1) 7.4.0")
Build machine cpu family: x86_64
Build machine cpu: x86_64
[...]

Issue #202, opened 2020-03-03
1.5.1 build fails without system install inih


ysblokje:
**Describe the bug**
Meson build generation fails due to typo in the inih dependency when no inih system package exists, this is with the the release archive NOT the git checkout of 1.5.1 that one actually works.

**To Reproduce**
./bootstrap.sh

**Expected behavior**
A complete build.

**System Info (please complete the following information):**
 - OS and version: Archlinux rolling
 - GameMode Version1.5.1

**Additional context**
Changing line 162 in the toplevel meson.buil from 

> fallback : ['inih', 'inih_dep']

to 

> fallback : ['inih', 'inih_dependency']

fixes the issue.

highvoltage:
Not sure if this is related, but I tried to build in debian with the system inih packages installed and got the following:

"""
/build/gamemode-1.5.1/obj-x86_64-linux-gnu/meson-private/tmp9ot471rp/testfile.c:14:17: note: each undeclared identifier is reported only once for each function it appears in

Checking for function "pidfd_open" : NO 
Configuring build-config.h using configuration
Pkg-config binary for MachineChoice.HOST is cached.
Determining dependency 'inih' with pkg-config executable '/usr/bin/pkg-config'
PKG_CONFIG_PATH: 
Called `/usr/bin/pkg-config --modversion inih` -> 1

CMake binary for MachineChoice.HOST is not cached
CMake binary missing from cross or native file, or env var undefined.
Trying a default CMake fallback at cmake
Trying CMake binary cmake for machine MachineChoice.HOST at [None]
Did not find CMake 'cmake'
Found CMake: NO
No CMake binary for machine MachineChoice.HOST not found. Giving up.
Run-time dependency inih found: NO (tried pkgconfig and cmake)
Looking for a fallback subproject for the dependency inih

meson.build:160:4: ERROR: Subproject exists but has no meson.build file
"""

afayaz-feral:
> this is with the the release archive NOT the git checkout of 1.5.1 that one actually works.
The line in the release archive seems to be identical to the one in the git tag.

'inih_dependency' was changed to 'inih_dep' in:

https://github.com/benhoyt/inih/commit/bf2f849c77e77479bef37e6cc4131e4164938316

@stephanlachnit any thoughts?

stephanlachnit:
> fixes the issue.

[...]

PR #207, opened 2020-03-13
Use wrap instead of git submodule

stephanlachnit:
I noticed that it's easier to use wrap (native meson dependency tool) than git submodules, since:
1. It doesn't require a git repository to work, source tree is enough.
2. It's less hassle to change or update.
3. It is a detached git head, so less annoying in git viewers.
4. Would have prevented #202.
5. Maybe more meson compatibility (#204).


stephanlachnit:
Closes #222 with the new r49 version.

PR #217, opened 2020-04-06
remove gitignore

stephanlachnit:
With meson gitignores are kinda useless, since the build files are de-coupled from the source code. Ignoring `build/` also is kinda sub-optimal since the meson project always uses `builddir` in their examples. Everyone can still use ignore stuff by using a local one .gitignore:
```gitignore
/.gitignore
/files/to/ignore
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,0 @@
-build/
-builddir/
-*.swp


gicmo:
Not completely useless since I still do `meson build` (I guess from CMake and old times in meson where the examples used [`build`](https://github.com/mesonbuild/meson/commit/f3be687cbbfd1ff25c7ec4ffe8ba4adfee9a6579). What is the harm of having a `.gitignore` file?

stephanlachnit:
> Not completely useless since I still do `meson build` (I guess from CMake and old times in meson where the examples used [`build`](https://github.com/mesonbuild/meson/commit/f3be687cbbfd1ff25c7ec4ffe8ba4adfee9a6579). What is the harm of having a `.gitignore` file?

IDE's can use the information in meson's builddir, ideally everyone would use the same dir, but as you see that is not the case. My IDE always wants to have the builddir for configuration (else it creates it, which is a bit annoying but not the point), but without a gitignores I always have to sort which files to stage and which not. It's not just the information in the builddir, you could also have a folder which stores the default build command for the IDE.
If everyone just uses their own gitignore, there is no need to create a giant gitignore with all the exceptions for every IDE. This is a bit more annoying when using make for example, but with meson every build file is in one single dir, so it doesn't even take a minute to set it up.

kakra:
You can always add local excludes for your private environment to `.git/info/exclude`. Really, your usecase should be handled via that file. This project uses `build`, and it ships with an appropriate `.gitignore`. If you want to handle additional exludes, `.git/info/exclude` is for you. :-)

stephanlachnit:
I'm aware of that, but it's just kinda stupid. Either remove the gitigonre file completely or at least include the build folder from meson's recommendation. Heck, meson even mentions on their wiki that gitignores are not really needed anymore: https://mesonbuild.com/Using-multiple-build-directories.html

I don't get the discussion about it, it's really simple to setup a local gitignore suited to your needs with meson.

stephanlachnit:
> I'm cool with it being removed, it's just a remnant of my solo development and was mostly useful for when users would primarily use the bootstrap script. It's served its purpose, kill it with fire.

Thanks for pointing this out. I pushed a commit to change the build folder in the bootstrap script and to remove build folder after installation.

kakra:
I'm not sure what just happened here... I'm fine with adding whatever build artifacts show up in my git status to my `.git/info/exclude`. That's what it is for. But I don't get the argument about why `.gitignore` should not exist in the repo at all? After all it's part of the repo and can exclude all the usual "don't want in the repo" files that could show up (like temporary files, caches, etc) for the devs convenience (or for security reasons, like excluding local configuration potentially holding passwords or other sensitive data). Maybe meson project follow some cool rules which makes it easy for IDEs to just hide the build stuff - and yes, I prefer that style of meson a lot, and don't get me wrong, that IS a cool feature. But that doesn't magically make `.gitignore` a file free to use for local purpose.

But that's probably off-topic for this project, it may be plausible that a `.gitignore` may not be needed here. I just wanted to point out that it's usually problematic to use it for local purpose, eventually at that step if a project introduced such a file later and you already had your local version in place. ;-)

stephanlachnit:
> But I don't get the argument about why .gitignore should not exist in the repo at all? After all it's part of the repo and can exclude all the usual "don't want in the repo" files that could show up (like temporary files, caches, etc) for the devs convenience

Yes, and this makes totally sense when a project use for example `make` and spams your folder with `.o` etc files - that's why it has been added to git in the first place. But ideally no project should need a gitignore.
Why? Image ignoring `build` like here, but then a contributor wants to add a file called `scripts/build/debug.sh` - it will be ignored, maybe even without it being noticed. You can solve this by setting it to `/build`, but then you can't read it with `tar`'s `--exclude-vcs-ignore` function (which isn't used here but easily could be in one of the scripts).
You can also read a discussion [here](https://github.com/libratbag/libratbag/pull/671).

> I just wanted to point out that it's usually problematic to use it for local purpose, eventually at that step if a project introduced such a file later and you already had your local version in place.

You are absolutely right about that! But I can also happen with the upstream gitignore the other way around.

Overall, yes this is very improbable here and I just could have changed the gitignore, but since it only was two lines long I found it better to just remove it.

afayaz-feral:
I'm in favour of removing this file, since the build is contained to a single folder.

Willing to keep it if it breaks #240, which looks like a really good solution to our current error prone method of manually updating the version.

stephanlachnit:
> Willing to keep it if it breaks #240, which looks like a really good solution to our current error prone method of manually updating the version.

I got an idea for #240 and this^^ The `--dirty` option there is only really useful when creating the release tarball anyway, when building it isn't.

PR #225, opened 2020-04-22
Add option to change lib dir in gamemoderun

### LINKED ISSUE #224: Ability to define gamemodeauto lib dir in gamemoderun
**Is your feature request related to a problem? Please describe.**
We currently ship our own gamemoderun as https://github.com/flathub/com.valvesoftware.Steam/blob/beta/resources/gamemoderun

because the way that the libdir for libgamemodeauto.so.0 is defined is not flexible enough for our sandbox layout. Specifically the usage of $LIB. Current setup is problematic as it may end up out of sync with gamemode ABI versioning.
**Describe the solution you'd like**
It would be nice to be able to define the entire library path as a single Meson variable. Current value is fine a default.
**Describe alternatives you've considered**
We could also patch gamemoderun but I'm not sure that would be much better than current workaround.

stephanlachnit:
Closes #224.

### DIFF
--- a/data/gamemoderun.in
+++ b/data/gamemoderun.in
@@ -3,7 +3,7 @@
 
 # Path to installed libgamemodeauto. ld.so will substitute "\$LIB" to get the
 # appropriate path depending on whether the app is 32- or 64-bit.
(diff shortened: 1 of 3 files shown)


jqadev:
@stephanlachnit is this PR ready to be marked as ready for review?

PR #227, opened 2020-04-23
Check index before using array element

qarmin:
`i` may be out of bounds if checking are used as second part of condition.


mdiluz:
Touché

Issue #222, opened 2020-04-16
libinih.so and libINIReader.so should be linked statically


kakra:
**Describe the bug**
When installing into a system without inih present, gamemode installs the headers and library as if they were their own package. Also, it installs pkgconfig files. This created a shadow/fake dependency. When later installing inih as a native system library, package managers will see file collisions.

Also, software may see inih pkgconfig as a dependency which is unrelated to gamemode.

Gentoo installation QA found this issue because in a multilib install, inih headers were present in the 64-bit install phase but not in the 32-bit install phase (Gentoo does not compile and install the tools and examples in 32-bit multilib just the daemon libs itself).

**To Reproduce**
1. Install gamemode without system inih present (using bundled lib)

**Expected behavior**
gamemode should not install inih libs system-wide, either it should link them statically or install them to a private lib directory. Also, the installed `gamemode_client.h` is totally unrelated to inih, so there's even no point in providing inih header and pkgconfig files.

I suggest the following files not being present in gamemode installed with bundled inih:

```
/usr/include/ini.h
/usr/include/INIReader.h
/usr/lib64/libinih.so.48
/usr/lib64/pkgconfig/INIReader.pc
/usr/lib64/pkgconfig/inih.pc
```

**System Info (please complete the following information):**
 - OS and version: Gentoo Linux
 - GameMode Version 1.5.1

**Additional context**
inih itself is unrelated to anything users or developers could do with `gamemode_client.h`, thus that dependency should be hidden from system linker view when bundled, either by linking statically or by placing in a private lib directory. I'd prefer static linking, the headers and pkgconfig have to be omitted to not collide with the package manager.

As a side note, meson should not fall back to bundled library automagically. It should always bundle it unless explicitly told otherwise. Or vice-versa and require explicit bundling if system-provided inih is missing. Usually, software bundling libs will always bundle them unless told to use the system-provided lib, so I'd prefer that variant.

stephanlachnit:
> When installing into a system without inih present, gamemode installs the headers and library as if they were their own package. Also, it installs pkgconfig files. This created a shadow/fake dependency. When later installing inih as a native system library, package managers will see file collisions.

That's true, it is not ideal at the moment as a fallback. I would recommend the Gentoo gamemode maintainer to just pack inih, then this doesn't happen.
[...]

PR #228, opened 2020-04-24
Add option to use elogind

### LINKED ISSUE #226: meson `-Dwith-systemd=false` is ignored
**Describe the bug**
Regardless of whether `--Dwith-systemd` is set false or true, it detects systemd:
```
# meson setup --buildtype plain --prefix /usr/local --wrap-mode nofallback -Dwith-systemd=false build
[...]
Found pkg-config: /usr/bin/pkg-config (0.29.2)
Run-time dependency libsystemd found: YES 244
[...]
```

On a non-systemd system, this leads to the following error because systemd support seems to be forced:
```
 * abi_x86_32.x86: running multilib-minimal_abi_src_configure
meson setup --buildtype plain --libdir lib --localstatedir /var/lib --prefix /usr --sysconfdir /etc --wrap-mode nodownload --cross-file /tmp/portage/games-util/gamemode-1.5.1/temp/meson.i686-pc-linux-gnu.x86 -Dwith-systemd=false -Dwith-systemd-user-unit-dir=/usr/lib/systemd/user -Dwith-examples=false -Dwith-daemon=false /tmp/portage/games-util/gamemode-1.5.1/work/gamemode-1.5.1 /tmp/portage/games-util/gamemode-1.5.1/work/gamemode-1.5.1-abi_x86_32.x86
...
...
Run-time dependency libsystemd found: NO (tried pkgconfig and cmake)
meson.build:83:0: ERROR: Dependency "libsystemd" not found, tried pkgconfig and cmake
```

See-also: https://github.com/gentoo/gentoo/pull/13158#discussion_r411892796

stephanlachnit:
Closes #226.

Didn't test it yet.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -79,9 +79,22 @@ path_libdir = join_paths(path_prefix, get_option('libdir'))
 path_libexecdir = join_paths(path_prefix, get_option('libexecdir'))
 path_sysconfdir = join_paths(path_prefix, get_option('sysconfdir'))
 
-# Find systemd via pkgconfig
+# Find systemd / elogind via pkgconfig
 with_systemd = get_option('with-systemd')
-dep_systemd = dependency('libsystemd')
+with_elogind = get_option('with-elogind')
+elogind_args = []
+dep_systemd = []
+if with_systemd == true
+    if with_elogind == true
+        error('you can\'t use systemd and elogind at the same time')
+    endif
+    dep_systemd = dependency('libsystemd')
+elif with_elogind == true
+    elogind_args += ['-DUSE_ELOGIND']
+    dep_systemd = dependency('libelogind')
+else
+    error('you need to use either sd-bus.h from systemd or elogind')
+endif
 
 # For the client, libdbus is used
 dep_dbus = dependency('dbus-1')
--- a/meson.build
+++ b/meson.build
@@ -80,20 +80,15 @@ path_libexecdir = join_paths(path_prefix, get_option('libexecdir'))
 path_sysconfdir = join_paths(path_prefix, get_option('sysconfdir'))
 
 # Find systemd / elogind via pkgconfig
-with_systemd = get_option('with-systemd')
-with_elogind = get_option('with-elogind')
-elogind_args = []
-dep_systemd = []
-if with_systemd == true
-    if with_elogind == true
-        error('you can\'t use systemd and elogind at the same time')
-    endif
-    dep_systemd = dependency('libsystemd')
-elif with_elogind == true
-    elogind_args += ['-DUSE_ELOGIND']
-    dep_systemd = dependency('libelogind')
-else
-    error('you need to use either sd-bus.h from systemd or elogind')
+sd_bus_provider = get_option('with-sd-bus-provider')
+
+sd_bus_args = []
+sd_bus_dep = []
+if sd_bus_provider == 'systemd'
+    sd_bus_dep = dependency('libsystemd')
+elif sd_bus_provider == 'elogind'
+    sd_bus_args += ['-DUSE_ELOGIND']
+    sd_bus_dep = dependency('libelogind')
 endif
 
 # For the client, libdbus is used
@@ -106,7 +101,7 @@ dep_threads = dependency('threads')
 libdl = cc.find_library('dl', required: false)
 
 # Determine the location for the systemd unit
-if with_systemd == true
+if sd_bus_provider == 'systemd'
     # If the path isn't explicitly set, ask systemd for the systemd user unit directory
     path_systemd_unit_dir = get_option('with-systemd-user-unit-dir')
     if path_systemd_unit_dir == ''
@@ -137,7 +132,6 @@ endif
 
 path_polkit_action_dir = join_paths(path_datadir, 'polkit-1', 'actions')
 
-with_daemon = get_option('with-daemon')
 with_examples = get_option('with-examples')
 with_util = get_option('with-util')
 
@@ -169,7 +163,7 @@ endif
 
 # The daemon can be disabled if necessary, allowing multilib builds of the
 # main library
-if with_daemon == true
+if sd_bus_provider != 'no-daemon'
     # inih currently only needed by the daemon
     inih_dependency = dependency(
         'inih',
@@ -200,7 +194,7 @@ report = [
     '    includedir:                             @0@'.format(path_includedir),
 ]
 
-if with_systemd == true
+if sd_bus_provider == 'systemd'
 report += [
     '    systemd user unit directory:            @0@'.format(path_systemd_unit_dir),
 ]
@@ -216,10 +210,9 @@ report += [
     '    Options:',
     '    ========',
     '',
-    '    daemon:                                 @0@'.format(with_daemon),
+    '    sd-bus provier:                         @0@'.format(sd_bus_provider),
     '    examples:                               @0@'.format(with_examples),
     '    util:                                   @0@'.format(with_util),
-    '    systemd:                                @0@'.format(with_systemd),
 ]
 
 # Output some stuff to validate the build config
--- a/daemon/gamemode-dbus.c
+++ b/daemon/gamemode-dbus.c
@@ -36,8 +36,13 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "common-logging.h"
 #include "common-pidfds.h"
 
+#ifdef USE_ELOGIND
+#include <elogind/sd-bus.h>
+#include <elogind/sd-daemon.h>
+#else
 #include <systemd/sd-bus.h>
 #include <systemd/sd-daemon.h>
+#endif
 
 #include <assert.h>
 #include <errno.h>
--- a/daemon/meson.build
+++ b/daemon/meson.build
@@ -17,11 +17,11 @@ gamemoded_includes += config_h_dir
 gamemoded = executable(
     'gamemoded',
     sources: daemon_sources,
-    c_args: elogind_args,
+    c_args: sd_bus_args,
     dependencies: [
         link_daemon_common,
         dep_threads,
-        dep_systemd,
+        sd_bus_dep,
         inih_dependency,
         libdl,
     ],
--- a/daemon/meson.build
+++ b/daemon/meson.build
@@ -17,6 +17,7 @@ gamemoded_includes += config_h_dir
 gamemoded = executable(
     'gamemoded',
     sources: daemon_sources,
+    c_args: elogind_args,
     dependencies: [
         link_daemon_common,
         dep_threads,
--- a/data/meson.build
+++ b/data/meson.build
@@ -18,7 +18,7 @@ config_example = run_command(
 ).stdout().strip()
 data_conf.set('GAMEMODE_EXAMPLE_CONFIG', config_example)
 
-if with_systemd == true
+if sd_bus_provider == 'systemd'
     # Install systemd user unit
     configure_file(
         input: 'gamemoded.service.in',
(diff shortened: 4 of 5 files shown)


kakra:
Compiles fine for systemd with the patch applied, still waiting for elogind results...

kakra:
Seems still to complain with the patch applied to 1.5.1:
```
>>> Configuring source in /tmp/portage/games-util/gamemode-1.5.1/work/gamemode-1.5.1 ...
 * abi_x86_32.x86: running multilib-minimal_abi_src_configure
meson setup --buildtype plain --libdir lib --localstatedir /var/lib --prefix /usr --sysconfdir /etc --wrap-mode nodownload --pkg-config-path=/usr/share/pkgconfig --cross-file /tmp/portage/games-util/gamemode-1.5.1/temp/meson.i686-pc-linux-gnu.x86 -Dsystemd=false -Delogind=true -Dwith-systemd-user-unit-dir=/usr/lib/systemd/user -Dwith-examples=false -Dwith-daemon=false --wrap-mode forcefallback /tmp/portage/games-util/gamemode-1.5.1/work/gamemode-1.5.1 /tmp/portage/games-util/gamemode-1.5.1/work/gamemode-1.5.1-abi_x86_32.x86
...
...
...
Run-time dependency libsystemd found: NO (tried pkgconfig and cmake)

meson.build:90:4: ERROR: Dependency "libsystemd" not found, tried pkgconfig and cmake
```

The meson cmdline has:

    -Dsystemd=false -Delogind=true -Dwith-systemd-user-unit-dir=/usr/lib/systemd/user

stephanlachnit:
Typo:
`-Dwith-systemd=false -Dwith-elogind=true`

kakra:
Oh, true... Gentoo doesn't add "with-" prefix to the feature name automatically...

kakra:
LGTM, works in theory but there's a multilib problem somewhere else. But that should be unrelated to this PR. So from my point of view it's ready for merge.

stephanlachnit:
> LGTM, works in theory but there's a multilib problem somewhere else. But that should be unrelated to this PR. So from my point of view it's ready for merge.

Same problem on Debian. It's not implemented yet properly, but we already split gamemode in 4 packages. This way the libraries are separated from the non-library files. If you want to have a look, [here](https://salsa.debian.org/stephanlachnit/gamemode_/-/tree/debian/master/debian) are Debian build files, the `xyz.install` files contain what is included in the package `xyz`.

stephanlachnit:
Also I noticed that this isn't the best approach in Meson to solve this. Instead it should be a `type : 'combo'`, where one can choose between `['systemd', 'elogind', 'no-dameon']`, replacing the `with-daemon` option as well. If I have time I will change it.

kakra:
Yes, a `no-daemon` option would be useful...

stephanlachnit:
Pushed as combo.

kakra:
Seems to close #222 for me: The abi32 compile pass no longer install the inih headers. Was this intentional?

stephanlachnit:
It wasn't intentional, so no idea

kakra:
That additional patch works for me. I'm good with it.

stephanlachnit:
Can you send me the build log?

kakra:
> Can you send me the build log?

Here's the complete log of the Gentoo ebuild, it builds for both abi64 and abi32, and it builds against systemd in my case:

[build.log](https://github.com/FeralInteractive/gamemode/files/4541805/build.log)

kakra:
Looking at the build log, there should probably be an option to build without the libexec tools but that should be a matter of another issue report. The build installs the 64 bit versions over the 32 bit versions, so all is fine in the final package image. Just a few wasted CPU cycles. In Gentoo, packages are not split between ABIs, they build and install both in the same package if asked to do so. The build sources for this package are here if you're interested: https://github.com/gentoo/gentoo/pull/13158

stephanlachnit:
> Here's the complete log of the Gentoo ebuild, it builds for both abi64 and abi32, and it builds against systemd in my case:

Something is wrong: `-Dwith-sd-bus-provider=systemd -Dwith-systemd-user-unit-dir=/usr/lib/systemd/user -Dwith-examples=false -Dwith-sd-bus-provider=no-daemon`

It's twice the same setting, and last one overrides, so the daemon isn't build at all, which means inih isn't builded at all.

kakra:
> It's twice the same setting, and last one overrides, so the daemon isn't build at all, which means inih isn't builded at all.

Yes but that's expected behavior in this case: When building multilib, it adds this is the abi32 case. So it's good it overwrites the first setting.

From the ebuild:

```bash
multilib_src_configure() {
        local emesonargs=(
                -Dwith-sd-bus-provider=$(usex systemd "systemd" "elogind")
                -Dwith-systemd-user-unit-dir="$(systemd_get_userunitdir)"
        )
        if ! multilib_is_native_abi; then
                emesonargs+=(
                        -Dwith-examples=false
                        -Dwith-sd-bus-provider=no-daemon
                )
        fi

        # Upstream bug: https://github.com/FeralInteractive/gamemode/issues/222
        emesonargs+=(
                --wrap-mode forcefallback
        )

        meson_src_configure
}
```

So if building for abi32 only (and thus it would be the native abi), those extra args are not added.

stephanlachnit:
Ah okay

kakra:
> Ah okay

But neither abi pass install inih headers now. So that's probably a side-effect of using forcefallback. I didn't closely watch the behavior change between my ebuild adjustments.

kakra:
There's a bug somewhere, from the build log:
```
 * abi_x86_64.amd64: running multilib-minimal_abi_src_configure
meson setup --buildtype plain --libdir lib64 --localstatedir /var/lib --prefix /usr --sysconfdir /etc --wrap-mode nodownload --pkg-config-path=/usr/lib64/pkgconfig -Dwith-sd-bus-provider=systemd -Dwith-systemd-user-unit-dir=/usr/lib/systemd/user --wrap-mode forcefallback /var/tmp/portage/games-util/gamemode-1.5.1/work/gamemode-1.5.1 /var/tmp/portage/games-util/gamemode-1.5.1/work/gamemode-1.5.1-abi_x86_64.amd64
```

In 64 bit build phase, it has `-Dwith-sd-bus-provider=systemd -Dwith-systemd-user-unit-dir=/usr/lib/systemd/user --wrap-mode forcefallback` and if I'm not blind, there's no other interfering option. But it doesn't build the daemon. It should, shouldn't it?

kakra:
Tried again in my local dev build:
```bash
git clean -xdf && meson build && ninja -C build
```

There's no daemon in the build directory.

stephanlachnit:
Maybe I messed something up, I'll look tmw.

meson.build
```diff
@@ -168,7 +162,7 @@ endif
 
 # The daemon can be disabled if necessary, allowing multilib builds of the
 # main library
-if with_daemon == true
+if sd_bus_provider == 'no-dameon'
```
kakra, line 165: This reverses the logic, doesn't it?

stephanlachnit: Ah yes thanks.

kakra:
Excerpt from the build log:
```
Run-time dependency dbus-1 found: YES 1.12.16
Run-time dependency threads found: YES
Library dl found: YES
Checking for function "pidfd_open" : NO
Configuring build-config.h using configuration
Looking for a fallback subproject for the dependency inih because:
Use of fallback dependencies is forced.

|Executing subproject inih method meson
|
|Project name: inih
|Project version: 48
|C compiler for the build machine: x86_64-pc-linux-gnu-gcc -m32 (gcc 9.3.0 "x86_64-pc-linux-gnu-gcc (Gentoo 9.3.0 p2) 9.3.0")
|C linker for the build machine: GNU ld.gold 2.33.1
|Appending CXXFLAGS from environment: '-O2 -march=native -pipe -fomit-frame-pointer -g -ftree-vectorize -ftree-slp-vectorize'
|Appending LDFLAGS from environment: ['-Wl,-O1', '-Wl,--as-needed']
|C++ compiler for the build machine: x86_64-pc-linux-gnu-g++ -m32 (gcc 9.3.0 "x86_64-pc-linux-gnu-g++ (Gentoo 9.3.0 p2) 9.3.0")
|C++ linker for the build machine: GNU ld.gold 2.33.1
|C compiler for the host machine: x86_64-pc-linux-gnu-gcc -m32 (gcc 9.3.0 "x86_64-pc-linux-gnu-gcc (Gentoo 9.3.0 p2) 9.3.0")
|C linker for the host machine: GNU ld.gold 2.33.1
|C++ compiler for the host machine: x86_64-pc-linux-gnu-g++ -m32 (gcc 9.3.0 "x86_64-pc-linux-gnu-g++ (Gentoo 9.3.0 p2) 9.3.0")
|C++ linker for the host machine: GNU ld.gold 2.33.1
|Build targets in project: 8
|Subproject inih finished.

Dependency inih from subproject subprojects/inih found: YES

data/meson.build:13:3: ERROR: Unknown variable "with_systemd".
```

stephanlachnit:
Tbh I'm not that happy with combining everything into one option anymore, unless I would rewrite the entire build files so it makes a little bit more sense with this option.

kakra:
Sorry:
```
data/meson.build:13:3: ERROR: Unknown variable "sd".
```

kakra:
I think it should use underscores:

    if sd-bus-prover == 'systemd'

kakra:
> Tbh I'm not that happy with combining everything into one option anymore, unless I would rewrite the entire build files so it makes a little bit more sense with this option.

I'm open to a different approach...

stephanlachnit:
> Sorry:

No need to apologize, I currently don't have my normal build system setup so I can't test it^^

kakra:
Here's another one:

    data/meson.build:13:3: ERROR: Unknown variable "sd_bus_prover".

Actually, I also missed to spot that. ;-)

stephanlachnit:
> Here's another one:

At this point it's just sad xd. I'm gonna invest more on time this when I have it, at least I hope all spelling mistakes are now sorted out.

kakra:
Yes, seems so. It doesn't link, tho. I appreciate your investment a lot.

I'll inspect this later, maybe clone your branch and fix what's missing.

No urge to work on this now but here's the problem:
```
# sd_bus_provider = "no-daemon"
daemon/f77b12a@@gamemoded@exe/gamemode-dbus.c.o:gamemode-dbus.c:game_vtable: error: undefined reference to 'sd_bus_object_vtable_format'
daemon/f77b12a@@gamemoded@exe/gamemode-dbus.c.o:gamemode-dbus.c:gamemode_vtable: error: undefined reference to 'sd_bus_object_vtable_format'
collect2: error: ld returned 1 exit status
ninja: build stopped: subcommand failed.
```

stephanlachnit:
Looks like it, no idea why though.

TheGreatMcPain:
There's a typo in meson.build on line 165.\
`if sd_bus_provider != 'no-dameon'` should be `if sd_bus_provider != 'no-daemon'`

Other than that, I was able to build it on Gentoo with elogind using [this ebuild.](https://github.com/TheGreatMcPain/TheGreatMcPain-overlay/blob/master/games-util/gamemode/gamemode-9999-r1.ebuild)

stephanlachnit:
> There's a typo in meson.build on line 165.

Ah thanks, that was it.

jqadev:
@stephanlachnit there appeared conflicts in `daemon/meson.build` that I guess prevent @afayaz-feral from doing a review.

stephanlachnit:
> @stephanlachnit there appeared conflicts in daemon/meson.build that I guess prevent @afayaz-feral from doing a review.

Thanks for the heads up, I'll just fix a compiler warning an then it should be ready to be reviewed.

Btw the second commit is optional, I reduces the number of options but it isn't exactly clearer, so any opinions are appreciated.

afayaz-feral:
> Btw the second commit is optional, I reduces the number of options but it isn't exactly clearer, so any opinions are appreciated.

I like that second commit; you wouldn't want multiple options at once so it makes sense to combine them.

Issue #214, opened 2020-03-28
gamemoded sets cpu governor to powersave after exiting


jug007:
**Describe the bug**
When exiting gamemode, gamemoded sometimes sets the cpu governor to powersave, even when performance was set before entering gamemode.

**To Reproduce**
Steps used to reproduce the behavior:
1. Check cpu governor (for example with `cpupower frequency-info`)
1. Launch gamemode (with `gamemoderun application`)
2. Exit application
3. View output of 'syslog'

**Expected behavior**
Governor should be set back to performance (or not changed in the first place as that is not needed)

**System Info (please complete the following information):**
 - OS and version: Linux Mint 19.3 (Ubuntu 18.04 based)
 - GameMode Version: v1.4

**Additional context**
Governor is set to powersave on system boot. I have no idea how to change that. I change it after boot with `sudo cpupower frequency-set --governor performance`. Driver is acpi-cpufreq.

Gamemode first start:
```
Mar 28 01:01:11 motofckr9k-mint-desktop /usr/bin/gamemoded[3980]: Adding game: 5529 [/bin/bash]
Mar 28 01:01:11 motofckr9k-mint-desktop /usr/bin/gamemoded[3980]: Entering Game Mode...
Mar 28 01:01:11 motofckr9k-mint-desktop /usr/bin/gamemoded[3980]: governor was initially set to [performance]
Mar 28 01:01:11 motofckr9k-mint-desktop /usr/bin/gamemoded[3980]: Requesting update of governor policy to performance
```

Gamemode exit (first and only one)
```
Mar 28 04:04:16 motofckr9k-mint-desktop /usr/bin/gamemoded[3980]: Removing game: 9748 [/bin/bash]
Mar 28 04:04:16 motofckr9k-mint-desktop /usr/bin/gamemoded[3980]: Leaving Game Mode...
Mar 28 04:04:16 motofckr9k-mint-desktop /usr/bin/gamemoded[3980]: Requesting update of governor policy to powersave
```

stephanlachnit:
Not a bug, this is actually intended. Check `/etc/gamemode.ini`: by default `defaultgov=powersave` is set, however if you want to have your default cpu governor to be performance, set it to `defaultgov=performance`.

Anjune:
[...]

PR #229, opened 2020-04-28
Make system revert to previous governor instead of "powersave"

Anjune:
Closes: #214


stephanlachnit:
If you write "Closes" before the issue, the issue will be automatically closed when merging the PR.

PR #237, opened 2020-05-11
extend and fix travis

stephanlachnit:
Currently travis fails for various reasons, this is an attempt to fix it (I'm still not entirely sure why it fails, I can't reproduce it). I also extended it: it now loads the example config and runs the example game to actually verify that is it working.

Btw, for some reason travis doesn't appear in PRs, I don't think this is intended (?)

Edit: looks like it works now: https://travis-ci.org/github/FeralInteractive/gamemode/builds/685589040

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ sudo: false
 addons:
   apt:
     packages:
+      - appstream
       - clang
       - clang-format
       - libinih-dev
@@ -18,6 +19,7 @@ addons:
 script:
   - ./scripts/format-check.sh
   - ./bootstrap.sh -Dwith-examples=true
+  - meson test -C build
   - gamemoded -v
   - dbus-run-session -- gamemode-simulate-game
   - ./scripts/static-analyser-check.sh 
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,5 @@ script:
   - ./scripts/format-check.sh
   - ./bootstrap.sh -Dwith-examples=true
   - meson test -C build
-  - gamemoded -v
   - dbus-run-session -- gamemode-simulate-game
   - ./scripts/static-analyser-check.sh 
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,9 @@ addons:
     packages:
       - clang
       - clang-format
-      - libsystemd-dev
+      - libinih-dev
       - libdbus-1-dev
+      - libsystemd-dev
       - meson
   artifacts:
     paths:
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,4 +18,5 @@ script:
   - ./scripts/format-check.sh
   - ./bootstrap.sh -Dwith-examples=true
   - gamemoded -v
+  - dbus-run-session -- gamemode-simulate-game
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -18,8 +18,7 @@ addons:
 
 before_script:
   - pip3 install wheel
-  - pip3 install --upgrade 'meson==0.52.0'
-  - meson --version
+  - pip3 install meson
```
afayaz-feral, line 21: Can we pin this to a specific version like before? i.e.

```suggestion
  - pip3 install --upgrade 'meson==0.54.1'
```

Using meson 0.53 caused Travis to fail without any changes on our part, and took some time to fix - time also potentially wasted assuming our new changes was the culprit.

It'd be best to avoid running into such issues again. I'd prefer updating the version after verifying it doesn't cause any issues.

Likewise, echoing the version info was useful in debugging the issue, so I'd like to keep it.

See 16982f7.

stephanlachnit: Updated, but I'm not sure if I like it that much. The problem in 16982f712633b9f612519a498d5c185fa2611ee2 was that we used a Meson version that wasn't compatible with bionic, it wasn't the fault of Meson. I agree that is annoying when the build fails because of that, but wouldn't then a better option be to just use the Meson version provided by the distro? Bionic has Meson 0.45.1, I don't know if that will still work with gamemode. Also possible would be to go with focal, and take it as a baseline.

mdiluz:
When I wrote this, Travis didn't have Bionix support yet, it does now but it won't have Focal for a long while: https://docs.travis-ci.com/user/reference/overview/#linux

.travis.yml
mdiluz: When I wrote this, Travis didn't have Bionix support yet, it does now but it won't have Focal for a long while: https://docs.travis-ci.com/user/reference/overview/#linux

afayaz-feral: If we're able to use the version of meson packaged with the distro I think that would be ideal. That will still avoid the problem of incompatible versions from pip but will be easier for us to maintain. Although if 0.45.1 ends up being too old, I'm not opposed to sticking to a pinned version from pip.

The main justification is to remove a potential unknown from our tests, regardless of the cause of the problem. I don't want the test environment to change without our knowledge, but always grabbing the latest version via pip has the potential to do exactly that.

stephanlachnit: > When I wrote this, Travis didn't have Bionic support yet, it does now but it won't have Focal for a long while: https://docs.travis-ci.com/user/reference/overview/#linux

I run travis CI instance with focal without any problems. It works, I guess it's just not advertised (yet).

stephanlachnit:
Okay, a couple of tries later I got it working with focal.
I also did more stuff:
- compile against system libinih, which is possible since focal
- validate the metainfo
- run `gamemoded -v` as a meson test

Travis log: https://travis-ci.org/github/FeralInteractive/gamemode/builds/686217341

afayaz-feral:
Looks great! Thanks for fixing it, and for the new additions.

I'm sure focal won't be going anywhere soon as it's an LTS release, so I'm not too fussed about it not being mentioned in the docs yet.

PR #238, opened 2020-05-11
remove some old example files

stephanlachnit:
- The desktop file isn't really necessary, gamemode is auto started anyway.
- The archlinux build files shouldn't really be in here, also they aren't really necessary anymore since gamemode now is available in a lot of distros, so there are enough places to look it up.


afayaz-feral:
Agreed, merged.

PR #239, opened 2020-05-12
Use distro provided git-clang-format

stephanlachnit:
While playing around with travis I noticed that it actually isn't necessary to download git-clang-format or to store it in this repo, since it is available in all major distributions. In Debian based distros, it's in the [clang-format](https://packages.debian.org/sid/amd64/clang-format/filelist) package, which already is marked to be installed in travis.


afayaz-feral:
Thanks a lot, I had it on my list to find a better solution. Similar to the problem I mentioned in #237 I didn't want to rely on downloading code that could change without warning - in this case, wget would often fail causing random failures.

stephanlachnit:
No changes, I only rebased to check the CI output after #237.

> Thanks a lot, I had it on my list to find a better solution. Similar to the problem I mentioned in #237 I didn't want to rely on downloading code that could change without warning - in this case, wget would often fail causing random failures.

Yes that's definitely something I agree with - I just stumbled by accident over the Debian package and saw that it is actually in there.

PR #240, opened 2020-05-12
Generate version via git instead of using -dev

stephanlachnit:
This is an attempt to remove the -dev commits after releases.
This does work quite nicely with distros too, since they either build unmodified from git anyway (e. g. Arch Linux), or from the source tarball, where no .git folder is given (e. g. Debian).

One thing I'm unsure is the `--dirty` option, this adds `-dirty` if there are non-committed changes in the workdir, it can be a nice reminder but also potentially break if your default build dir is not in a gitignore (see #217), so I would like to hear some opinions here.

### DIFF
--- a/scripts/mkrelease.sh
+++ b/scripts/mkrelease.sh
@@ -7,10 +7,9 @@ set -e
 meson subprojects download
 meson subprojects update
 
-# Bump in tandem with meson.build, run script once new tag is up.
-VERSION="1.6-dev"
-
 NAME="gamemode"
+VERSION=$(git describe --tags --dirty)
+
 ./scripts/git-archive-all.sh --format tar --prefix ${NAME}-${VERSION}/ --verbose -t HEAD ${NAME}-${VERSION}.tar
 xz -9 "${NAME}-${VERSION}.tar"
 


afayaz-feral:
Perhaps we could pass an argument to mkrelease.sh with the path of the build dir we originally used? Then we could see about getting `git describe` and the archive script to ignore that path?

stephanlachnit:
> Perhaps we could pass an argument to mkrelease.sh with the path of the build dir we originally used? Then we could see about getting `git describe` and the archive script to ignore that path?

Now that I look at it again - `mkrelease.sh` calls `git-archive-all.sh`, which should only load files tracked by git. Since there is no submodule anymore, this file need to be re-written anyway. Working on it rn, will probably come back with a PR soon.

kakra:
Relying on Git only to create the version string will break builds in Gentoo as Gentoo downloads the tar.gz from the releases page, and that doesn't contain Git metadata. This will probably apply to other distro build scripts, also. So at least there should be a fallback to specify the version via meson option (and it that case not use Git). Alternatively, `mkrelease.sh` could create a version number to be contained in the final tar.gz.

Currently, it seems `gamemode_version` is still hard-coded and not dynamically parsed from Git at build time. So all should be fine. I just wanted to note that the build scripts may be run without any `.git` dir existing.

stephanlachnit:
The same applies to Debian. I have to rework this, but it is possible to set `meson.project_version()` as a fallback.

kakra:
It can be possible to completely omit version strings from the source and relying on tags only. A distro build script would just pass in the version string by synthesizing it from its own package version. If not passed as an option, the source build could fall back to using `git --describe`. I've done something similar for `bees` and it works great.

In Gentoo, we have a special package version 9999 (or anything else, some use 99999) do indicate a build from live Git. In that case, the ebuild script would see a `.git` directory and can omit passing the version to the source build script, otherwise I just pass the package version to the source build script and all is fine. It's a simple `if $PV != 9999 then add_the_version_option()` (in pseudo code).

What do you think?

stephanlachnit:
Actually there more I think about it, I don't think there is a real advantage by setting the version to something higher when using the git master branch, i. e. whether it's `1.5.1`, `1.6-dev` or `1.5.1-29-gf9eba3d`, it doesn't really make a difference in this case (not speaking of distro versioning ofc).
I would love to just use the `vcs_tag()` function from Meson, but it's only works for C headers apparently.
I think it's enough to use `git describe --tags --dirty` in `mkrelease.sh` and don't change the Meson project version after a release. This way it also can't interfere with any distro build system.

stephanlachnit:
@afayaz-feral I think the merge messed it up somehow. https://github.com/FeralInteractive/gamemode/blob/master/scripts/mkrelease.sh#L8 isn't generated by git.

afayaz-feral:
Definitely looks like a conflict issue that I missed, thanks for the heads up!

PR #253, opened 2020-06-03
README.md: Update note for Nvidia Optimus proprietary driver users

ahsanhussain:
Add a note for Nvidia hybrid graphics users, not having the opensource
nouveau driver, which can save them some time for searching the right
solution

Signed-off-by: Ahsan Hussain <ahsan_hussain@mentor.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ Config files are loaded and merged from the following directories, in order:
 
 It's not possible to integrate commands like optirun automatically inside GameMode, since the GameMode request is made once the game has already started. However it is possible to use a hybrid GPU wrapper like optirun by starting the game with `gamemoderun`.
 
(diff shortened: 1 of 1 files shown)


stephanlachnit:
Good idea, however I would remove `optirun` then since it isn't really a good solution anymore (either use nouveau DRI or nvidia's glvnd DRI).

ahsanhussain:
@stephanlachnit people using old distros/GPU drivers may still be using bumblebee, so shouldn't we keep it for them?

afayaz-feral:
Thanks for the update! I'm happy to leave the `optirun` example in, since it's just that - an example.

I like that it'll give people a good starting point to search for more info on the most appropriate option, if they're not already aware of it.

PR #256, opened 2020-06-16
Fix gamemoded --test

stephanlachnit:
That did the trick for me, let's see if it works on the CI as well.


stephanlachnit:
@afayaz-feral can you enable pull request reviews for travis? Somehow it doesn't pick it up, and also it's a bit annoying to always open the travis webpage.

stephanlachnit:
ok nvm, this will never be fixed since travis doesn't allow to set the cpu gov.

PR #262, opened 2020-07-05
Refactor attempt while providing upstream downstream

Kreyren:
DO_NOT_MERGE: Work in progress


Kreyren:
@FeralInteractive Please provide upstream-way of building (ideally with explanation) for this binary so that i can make a proper logic, thank you ^-^

mdiluz:
I'm not sure I understand the goals here or what's missing in the current repo - could you explain a little more?

Kreyren:
@mdiluz I want to change following
- `bootstrap.sh`, because it's not passing shellcheck and is seemingly designed to work only on specified systems with lack of sanitization
- Add CI to capture these code quality issues
- openrc support so that i can use this on my system (which doesn't seem to be supported atm, but seems to be easily implemented?)

I would also like to add exherbo and devuan downstream to upstream if you want it otherwise i will keep it in my fork, because keeping this on upstream repos seems to make it easier to work with. ^-^

Assuming it being welcomed?

Makefile
```diff
@@ -0,0 +1,25 @@
...
+				&& usermod -a -G tty paludisbuild \
+				&& printf '%s\n' \
+					\"format = e\" \
+					\"location = \"\$$\{root\}/var/db/paludis/repositories/gamemode\"\" \
+					\"sync = file:///var/repositories/gamemode\" \
+				> /etc/paludis/repositories/gamemode.conf \
+				&& cat /etc/paludis/repositories/gamemode.conf \
+				&& cave sync \
+				&& cave resolve games-utils/gamemode \
+			"
```
stephanlachnit, line 25: Adding a second build system (Make) just for a test that could be run in a script is a bad idea, it ruins automatic build systems like debhelper on Debian.

dist/README.md
```diff
@@ -0,0 +1,3 @@
+# Downstream management
+
+Directory dedicated for various package managers to recognize this repository into a package
```
stephanlachnit, line 3: WTF no please don't do something like this upstream. Package files should be downstream, always. It only bloats the repo and you can't even directly change something in there since you don't have write rights.

Makefile
stephanlachnit, line 25: Adding a second build system (Make) just for a test that could be run in a script is a bad idea, it ruins automatic build systems like debhelper on Debian.

bootstrap.sh
```diff
@@ -1,48 +1,383 @@
-#!/bin/bash
-# Simple bootstrap script to build and run the daemon
+#!/bin/sh
+# Originally created by FeralInteractive <https://www.feralinteractive.com/en/> under BSD 3-Clause License <https://github.com/FeralInteractive/gamemode/blob/master/LICENSE.txt> in 2017 bootstrap.sh
+# Performed full refact by Jacob Hrbek <kreyren@rixotstudio.cz> under GPLv3 license <https://www.gnu.org/licenses/gpl-3.0.en.html> in 05.07.2020 04:28:55 CET
```
stephanlachnit, line 3: Ignoring the need for this, please don't put license clutter in here. The file was already licensed under a BSD license, so no need to change that to a GPL license.

bootstrap.sh
```diff
@@ -1,48 +1,383 @@
...
+###! - [ ] FreeBSD
+###! - [ ] Darwin
+###! - [ ] Redox
+###! - [ ] ReactOS
+###! - [ ] Windows
+###! - [ ] Windows/Cygwin
+
+# Maintainer info
+UPSTREAM="https://github.com/FeralInteractive"
+MAINTAINER="https://github.com/Kreyren"
```
stephanlachnit, line 31: It's nice that you want to maintain this file, but I think that's something you should ask beforehand since you don't have write access to this repo.

bootstrap.sh
```diff
@@ -1,48 +1,383 @@
...
+	case "$unameKernel" in
+		Linux)
+			KERNEL="$unameKernel"
 
-# Restart polkit so we don't get pop-ups whenever we pkexec
-if systemctl list-unit-files |grep -q polkit.service; then
-    sudo systemctl try-restart polkit
+			# Assume Linux Distro and release
+			# NOTICE(Krey): We are expecting this to return a lowercase value
+			if command -v "$LSB_RELEASE" 1>/dev/null; then
```
stephanlachnit, line 280: I don't think it is necessary to identify either the kernel or the distro

bootstrap.sh
```diff
@@ -1,48 +1,383 @@
...
+				;;
+				*) die fixme "Unexpected Linux distribution '$assumedDistro' has been detected."
+			esac
+
+			# Verify Linux Distro Release
+			efixme "Sanitize verification of linux distro release"
+			assumedRelease="$RELEASE"
+		;;
+		FreeBSD | Redox | Darwin | Windows)
+			KERNEL="$unameKernel"
```
stephanlachnit, line 309: Useless, gamemode won't run on any of those since they don't have systemd or even elogind and probably will never have.

bootstrap.sh
```diff
@@ -1,48 +1,383 @@
...
+[ -z "$CHMOD" ] && CHMOD="chmod"
+[ -z "$UNAME" ] && UNAME="uname"
+[ -z "$TR" ] && TR="tr"
+[ -z "$SED" ] && SED="sed"
+[ -z "$GREP" ] && GREP="grep"
+[ -z "$DPKG" ] && DPKG="dpkg"
+[ -z "$APT_GET" ] && APT_GET="apt-get"
+[ -z "$TOR" ] && TOR="tor"
+[ -z "$NINJA" ] && NINJA="ninja"
+[ -z "$MESON" ] && MESON="meson"
```
stephanlachnit, line 47: Useless.

bootstrap.sh
```diff
@@ -1,48 +1,383 @@
...
+### Fixme trap
+[ -z "$DIE_FORMAT_STRING_FIXME" ] && DIE_FORMAT_STRING_FIXME="FATAL: %s in script '$myName' located at '$0', fixme?\n"
+[ -z "$DIE_FORMAT_STRING_FIXME_LOG" ] && DIE_FORMAT_STRING_FIXME_LOG="${logPrefix}FATAL: %s, fixme?\n"
+[ -z "$DIE_FORMAT_STRING_FIXME_DEBUG" ] && DIE_FORMAT_STRING_FIXME_DEBUG="FATAL($myName:$1): %s, fixme?\n"
+[ -z "$DIE_FORMAT_STRING_FIXME_DEBUG_LOG" ] && DIE_FORMAT_STRING_FIXME_DEBUG_LOG="${logPrefix}FATAL($myName:$1): %s, fixme?\\n"
+### Unexpected trap
+[ -z "$DIE_FORMAT_STRING_UNEXPECTED" ] && DIE_FORMAT_STRING_UNEXPECTED="FATAL: Unexpected happend while %s in $myName located at $0\\n"
+[ -z "$DIE_FORMAT_STRING_UNEXPECTED_LOG" ] && DIE_FORMAT_STRING_UNEXPECTED_LOG="${logPrefix}FATAL: Unexpected happend while %s\\n"
+[ -z "$DIE_FORMAT_STRING_UNEXPECTED_DEBUG" ] && DIE_FORMAT_STRING_UNEXPECTED_DEBUG="FATAL($myName:$1): Unexpected happend while %s in $myName located at $0\\n"
+[ -z "$DIE_FORMAT_STRING_UNEXPECTED_DEBUG_LOG" ] && DIE_FORMAT_STRING_UNEXPECTED_DEBUG="${logPrefix}FATAL($myName:$1): Unexpected happend while %s\\n"
```
stephanlachnit, line 84: What is this?

bootstrap.sh
```diff
@@ -1,48 +1,383 @@
...
+myName="${0##*/}"
+
+# Used to prefix logs with timestemps, uses ISO 8601 by default
+logPrefix="[ $(date -u +"%Y-%m-%dT%H:%M:%SZ") ] "
+# Path to which we will save logs
+# NOTICE(Krey): To avoid storing file '$HOME/.some-name.sh.log' we are stripping the '.sh' here
+logPath="${XDG_DATA_HOME:-$HOME/.local/share}/${myName%%.sh}.log"
+
+# inicialize the script in logs
+"$PRINTF" '%s\n' "Started $myName on $("$UNAME" -s) at $(date -u +"%Y-%m-%dT%H:%M:%SZ")" >> "$logPath"
```
stephanlachnit, line 99: Honestly this is so small there is no need to log this. Users can simply copy the output.

.github/workflows/shell.yml
```diff
@@ -0,0 +1,66 @@
...
+        shell: bash
+        run: |
+          cd "$GITHUB_WORKSPACE"
+
+          # Process files
+          ## NOTICE: Do not use for loop to avoid pitfall https://mywiki.wooledge.org/BashPitfalls#pf1
+          git --git-dir="$GITHUB_WORKSPACE/.git" ls-files -z -- '*.sh' | while IFS= read -rd '' file; do
+              printf 'linting shell file %s\n' "$file"
+              shellcheck --external-sources --shell=sh "$file"
+          done
```
stephanlachnit, line 66: Besides a spellcheck being pretty much useless wasted computation time, this can be easily integrated into travis in like two lines.

.github/workflows/exherbo-build.yml
```diff
@@ -0,0 +1,20 @@
...
+  # Linting
+  lint:
+    runs-on: ubuntu-latest
+    container: exherbo:exherbo-x86_64-pc-linux-gnu-base
+    steps:
+      - name: FIXME
+        # Make sure that bash is used
+        shell: bash
+        run: |
+          printf '%s\n' "FIXME"
```
stephanlachnit, line 20: I don't want to downplay Exherbo - but before a CI for Exherbo is added, one might add Fedora or Arch first since they have a much bigger user and developer base, and thus are more representative if the CI would actually fail.

.github/workflows/bootstrap-test.yaml
```diff
@@ -0,0 +1,20 @@
...
+  # Linting
+  lint:
+    runs-on: ubuntu-latest
+    container: debian:stable
+    steps:
+      - name: FIXME
+        # Make sure that bash is used
+        shell: bash
+        run: |
+          printf '%s\n' "FIXME"
```
stephanlachnit, line 20: Already included into travis, no need for this.

.github/workflows/bash.yml
```diff
@@ -0,0 +1,67 @@
...
+              exit 255
+            fi
+            # Install git
+            apt install -y git
+          elif apt list --installed 2>/dev/null | grep -qP "^git\/stable.*"; then
+            true
+          else
+            exit 255
+          fi
+          # Check for shellcheck
```
stephanlachnit, line 37: No need to run the spellcheck twice.

stephanlachnit:
> Currently it is painful to build this project due to the lack of code quality assurance in bootstrap.sh and seemingly hard dependency on systemd

First of all, I don't think that there is a lack of "QA" for this tiny bootstrap script. And gamemode it doesn't have a hard dependency on systemd either. It does require `libsystemd`, aka `sd-bus`, which btw doesn't require to have systemd as an init system.
Anyway, you probably want to use elogind for this, and there is a Meson build option for this.

> openrc support so that i can use this on my system (which doesn't seem to be supported atm, but seems to be easily implemented?)

So in this case, you might not want to run `bootstrap.sh` but rather write a custom script just for that purpose. I guess it doesn't make much sense to write a build script covering every possible option, so it only works for the majority of people. Distributions like Debian for example don't even use that script since they simply build with Meson, and set up the rest themselves.
Basically you want to do
```sh
meson builddir -Dwith-sd-bus-provider=elogind --prefix=/usr
meson compile -C builddir
sudo meson install -C builddir
```
And restart polkit somehow. That's it, no need for 400+ loc. Put that in your build files and distribute the package.

mdiluz:
I largely agree with @stephanlachnit here, there's no reason to put downstream stuff into the upstream repo. I also want to comment a little:

> bootstrap.sh, because it's not passing shellcheck and is seemingly designed to work only on specified systems with lack of sanitization

This is a reason to delete it, not add 500 lines of code.

And maybe this is simply a reminder to move the travis stuff over to github actions now that those are allowed for free.

kakra:
I guess it should simply be pointed out that `bootstrap.sh` is not to be used for packaging. I usually avoid such scripts even for a simple one-shot installation directly from source, because such scripts often make wrong assumptions about my system, i.e. they use sudo to clutter and clobber my system, overwriting files in `/usr/lib` or simply putting files not tracked by the package manager all over the system folders.

In this instance, it will install into `/usr` prefix by default. That should be `/opt/gamemode` or `/usr/local` probably. I'd prefer one-shot installations to live in their own namespace below `/opt` so I could uninstall by just deleting it. An unmanaged installer should never install to `/usr` directly.

As such, the script should probably be removed, or the defaults should change. In short term, the install instruction should put a clear message that `bootstrap.sh` is not to be used by packagers, maybe by introducing a new "Packaging" section in the README file showing how to use meson/ninja and how to manage the systemd/elogind dependencies (these are not properly documented yet, this is the opportunity).

stephanlachnit:
> I guess it should simply be pointed out that `bootstrap.sh` is not to be used for packaging. I usually avoid such scripts even for a simple one-shot installation directly from source, because such scripts often make wrong assumptions about my system, i.e. they use sudo to clutter and clobber my system, overwriting files in `/usr/lib` or simply putting files not tracked by the package manager all over the system folders.
> 
> In this instance, it will install into `/usr` prefix by default. That should be `/opt/gamemode` or `/usr/local` probably. I'd prefer one-shot installations to live in their own namespace below `/opt` so I could uninstall by just deleting it. An unmanaged installer should never install to `/usr` directly.
> 
> As such, the script should probably be removed, or the defaults should change. In short term, the install instruction should put a clear message that `bootstrap.sh` is not to be used by packagers, maybe by introducing a new "Packaging" section in the README file showing how to use meson/ninja and how to manage the systemd/elogind dependencies (these are not properly documented yet, this is the opportunity).

Yeah I guess it's a good idea to make it more clear that the scripts is only intended if you install from source. Switching to `/usr/local` is probably a good idea, but it will break `gamemoded  --test` because that path is hardcoded, something that should be fixed for that.

PR #263, opened 2020-07-06
simplify libgamemodeauto preloading

### LINKED ISSUE #224: Ability to define gamemodeauto lib dir in gamemoderun

stephanlachnit:
I've recently been working on MangoHud and setting up a 32 bit library preloading script similar to what we have here, and I noticed that we don't actually need the changes made in #225.
The PR was submitted to fix #224, however it was overkill for that and is also a little bit messy. Instead of defining the path to the shared library, it's better to just don't define it at all. `ld` will look it up in the systems search path, and will also pick the right one depending on whether it's a 32- or 64-bit app.
For the Flathub package this is ideal (/cc @nanonyme), since they can define the library lookup path without the need to mess around with the script. That this is a better solution will be clear when looking at their launch script: https://github.com/flathub/com.valvesoftware.Steam/blob/beta/resources/gamemoderun, it is based on the same principle.


nanonyme:
Oops, sorry, I was intending to have discussion on whether there was a reason not to trust dynamic linker to give you a gamemodeauto library. Thanks for bringing this up.

nanonyme:
But yes, for general-purpose app doing what com.valvesoftware.Steam does may be a bit involved. That app needs the interesting stunts with links and $LIB anyway because it uses an audit library and those need full path to library. The suggestion for earlier fix was based on assumption there's some security or whatnot concern of using full path which may or may not be true. It would be nice to have feedback from project maintainers on that regard.

stephanlachnit:
> The suggestion for earlier fix was based on assumption there's some security or whatnot concern of using full path which may or may not be true. It would be nice to have feedback from project maintainers on that regard.

The original preload was written from @mdiluz, back then as `/usr/lib/libgamemodeauto.so`. I guess there was no big reasoning behind using the full path vs just the file name since it didn't matter without Mutli-Arch anyway.

mdiluz:
It's been so long I can't remember why I specified the full path, perhaps because I originally didn't have it installed within the dynamic linker's normal search paths. This seems sane

PR #269, opened 2020-08-11
Align open_fdinfo_dir with pidfds_to_pids

nthykier:
The function `pidfds_to_pids` expected `-1` if `open_fdinfo_dir` failed but the latter returned `errno` which is hard to distinguish from a valid file handle.  Correct that by making `open_fdinfo_dir` a wrapper around `open`.


mdiluz:
Nice spot, LGTM

Issue #325, opened 2021-05-24
Move to github actions


stephanlachnit:
Travis is failing for no apparent reason (or to be more precise, their apt addon is broken), so I suggest to move to GitHub actions. I can provide a PR if you are fine with this change @afayaz-feral.

It's relatively easy to do, we can take this as a base: https://github.com/stephanlachnit/yaiden/blob/main/.github/workflows/build-and-test.yml

afayaz-feral:
I'm happy to make this change, I'll test and approve the PR if you make one.

I was just about to delve into what's going on here so thank you for saving me the trouble. I will hold off approving the other open PRs until this is resolved.

PR #326, opened 2021-05-24
Move to GitHub Actions

stephanlachnit:
Closes https://github.com/FeralInteractive/gamemode/issues/325.


stephanlachnit:
Run: https://github.com/stephanlachnit/gamemode/actions/runs/872646846

stephanlachnit:
@afayaz-feral This should be good to merge now. Here is the output of the run: https://github.com/stephanlachnit/gamemode/runs/2660255091

I tested the format check with a negative example, and it fails as expected. But I had to adjust the script, using `git-clang-format` would somehow fail. Game simulation also works, and so does the static analyser.

As you can see, the script currently fails. Since it runs at the end, I would still merge this and fix the issues found later (better a CI that fails at the end than one that doesn't even start properly).

stephanlachnit:
Maybe relevant for the static analyser: https://clang-analyzer.llvm.org/faq.html#null_pointer

afayaz-feral:
> better a CI that fails at the end than one that doesn't even start properly

Agreed. I'll merge this now - apologies for the delay.

PR #339, opened 2021-09-20
data: add and install sysusers.d gamemode.conf

evelikov:
Add a trivial gamemode.conf file, which creates the gamemode group.

### DIFF
--- a/data/gamemode.conf
+++ b/data/gamemode.conf
@@ -1,0 +1,1 @@
+g gamemode - -
--- a/data/meson.build
+++ b/data/meson.build
@@ -19,6 +19,11 @@ if sd_bus_provider == 'systemd'
         configuration: data_conf,
         install_dir: path_systemd_unit_dir,
     )
+    # Install the sysusers.d file
+    install_data(
+        files('gamemode.conf'),
+        install_dir: path_systemd_sysusers_dir,
+    )
 endif
 
 # Install the D-BUS service file
--- a/meson.build
+++ b/meson.build
@@ -111,6 +111,8 @@ if sd_bus_provider == 'systemd'
         pkgconfig_systemd = dependency('systemd')
         path_systemd_unit_dir = pkgconfig_systemd.get_pkgconfig_variable('systemduserunitdir')
     endif
+    pkgconfig_systemd = dependency('systemd')
+    path_systemd_sysusers_dir = pkgconfig_systemd.get_pkgconfig_variable('sysusersdir')
 endif
 
 with_limits_conf = get_option('with-pam-group')


evelikov-work:
@mdiluz are you still involved in gamemode, any comments?

mdiluz:
I am, I'll chase up with some contacts at Feral. Might be worth explaining why this is needed though.

evelikov-work:
Thanks everyone. To elaborate on the reason why, I think this is a good idea:

AFAICT gamemode manages sysfs entries, where most of the setters (cpu_gov, amdgpu dpm) require root permissions. Using root isn't particularly good, since it exposes much larger surface for misuse or abuse. To avoid that, we add a dedicated group, annotate respective entries and ultimately tighten the scope.

kira-bruneau:
On NixOS, system configuration is handled separately from packaging, and the package fails to build with this change: NixOS/nixpkgs#182511.

```
PermissionError: [Errno 13] Permission denied: '/nix/store/fi2ksazpsqkvspqf36ivlp5ghpxbfm9i-systemd-250.4/lib/sysusers.d'
FAILED: meson-install
/nix/store/1dbfsvzd4i8kdc3yb081gpfddlb80858-meson-0.61.2/bin/meson install --no-rebuild
ninja: build stopped: subcommand failed.
```

1. Should a flag be added to `meson_options.txt` to make this optional?
2. Where exactly is this group used, and how does it differ than what can be specified for `with-pam-group`?

evelikov:
AFAICT the file must be installed in `pkgconfig_systemd.get_pkgconfig_variable('sysusersdir')` in order for systemd to pick it up.

I don't know much about the working of NixOS, but it does sound alarming to use the default config provided by upstream. Nevertheless - feel free to add a meson toggle to disable it. There's systemd-user-unit-dir in there already.

kira-bruneau:
Nix does support shipping default configs provided from upstream, as long as the package installs to the `$out` directory specified in the build (eg. `$out/share/systemd/user/gamemoded.service`).

The problem here is that `pkgconfig_systemd.get_pkgconfig_variable('sysusersdir')` returns a path to an unwritable directory outside of `$out` (`/nix/store/fi2ksazpsqkvspqf36ivlp5ghpxbfm9i-systemd-250.4/lib/sysusers.d`), which is why the build fails.

evelikov-work:
Yes that was fairly obvious by the `Permission denied` message itself. What I'm saying is that, systemd modules (like sysusers but not limited to) will look into the folder specified in systemd.pc.

If the files are installed in that location, then the module will not work as expeted.

It sounds to me that if a feature (again sysusers or otherwise) is disabled, then the respective path should either be a) unset in the systemd.pc or b) pointing to special location that will not cause issues.

Either way, as mentioned originally feel free to add meson toggle if that makes sense for NixOS.

PR #356, opened 2022-02-26
Fix GitHub Actions Static Analyzer Check

samrocketman:
Dockerfile
----------

This dockerfile was used to emulate GitHub actions.

<details><summary>Dockerfile source (click to expand)</summary>

```dockerfile
FROM ubuntu:20.04

ENV DEBIAN_FRONTEND=noninteractive

RUN set -ex; \
apt-get update; \
apt-get install -y build-essential meson appstream clang clang-format clang-tools libdbus-1-dev libinih-dev libsystemd-dev git

RUN set -ex; \
yes | adduser ci-user

USER ci-user
```

</details>

Commands to reproduce
---------------------

I simulated GitHub actions with a local docker environment.

    docker build -t ci .

I ran the following commands to reproduce the issue so that I could bugfix the static analyzer script.

```bash
docker run -e CI=true --rm -v "$PWD:$PWD" -w "$PWD" --init ci ./scripts/static-analyser-check.sh
docker run -e CI=true --rm -v "$PWD:$PWD" -w "$PWD" --init ci ./bootstrap.sh -Dwith-examples=true
docker run --rm -v "$PWD:$PWD" -w "$PWD" --init ci meson test -C builddir
docker run --rm -v "$PWD:$PWD" -w "$PWD" --init ci dbus-run-session -- gamemode-simulate-game
docker run --rm -v "$PWD:$PWD" -w "$PWD" --init ci ./scripts/static-analyser-check.sh
```

All commands reuse the same workspace which is how it works in GH actions.

Legitimate bugs
---------------

Now the error displayed in build failure appears to be a legitimate bug.  Four were found related to null pointer dereference.  I'm not fixing the bug but the check should work properly for those who do.

If you following my docker setup in this description you can view the bug report by starting the report server.

```bash
docker run -p 127.0.0.1:8181:8181 -it --rm -v "$PWD:$PWD" -w "$PWD" --init ci scan-view --host 0.0.0.0 --allow-all-hosts builddir/meson-logs/scanbuild/*
```

After the report server is started you can visit `http://127.0.0.1:8181/` to view the bugs.

Screenshot of report
--------------------

Screenshot of bug report

# Le petit chat cool

image

### DIFF
--- a/scripts/static-analyser-check.sh
+++ b/scripts/static-analyser-check.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -exo pipefail
+
 # Ensure we are at the project root
 cd "$(dirname $0)"/..
 
@@ -7,4 +9,4 @@ cd "$(dirname $0)"/..
 ninja scan-build -C builddir | tee builddir/meson-logs/scan-build.txt
 
 # Invert the output - if this string exists it's a fail
-exit ! grep -E '[0-9]+ bugs? found.' builddir/meson-logs/scan-build.txt
+! grep -E '[0-9]+ bugs? found.' builddir/meson-logs/scan-build.txt
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+/builddir


scripts/static-analyser-check.sh
```diff
@@ -1,10 +1,12 @@
...
+
 # Ensure we are at the project root
 cd "$(dirname $0)"/..
 
 # Collect scan-build output
 ninja scan-build -C builddir | tee builddir/meson-logs/scan-build.txt
 
 # Invert the output - if this string exists it's a fail
-exit ! grep -E '[0-9]+ bugs? found.' builddir/meson-logs/scan-build.txt
+! grep -E '[0-9]+ bugs? found.' builddir/meson-logs/scan-build.txt
```
samrocketman, line 12: `exit` only allows integers.  However, it doesn't make sense to use exit anyways since bash supports inverting exit codes.

scripts/static-analyser-check.sh
```diff
@@ -1,10 +1,12 @@
 #!/bin/bash
 
+set -exo pipefail
```
samrocketman, line 3: https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html

* `set -e` exit immediately if any command line fails.
* `set -x` print the command to stderr and then execute it (helps know which commands fail and where).
* `set -o pipefail` exit script when any command in a piped sequence fails.  Example: `a | b | c` if command `a` or `b` fails the script would not exit unless this option is set.

samrocketman:
# Look at the CI logs

If you visit GitHub actions failure in this PR you'll see it fails properly for a legitimate reason (this should be merged even failing and a followup PR could be made to fix bugs).

    scan-build: 4 bugs found.

https://github.com/FeralInteractive/gamemode/runs/5272102252?check_suite_focus=true is an example of the error without the fix provided in this pull request.  You get an exit integer error instead of a legitimate report for scan errors.

    ./scripts/static-analyser-check.sh: line 10: exit: !: numeric argument required

afayaz-feral:
Looks good to me. The errors seem to be simple `NULL` checks.

PR #370, opened 2022-07-10
ci: fix static analyser with assertions

stephanlachnit:
@afayaz-feral this should fix the CI.

The reason it failed is in `lib/gamemode_client.h`:
screenshot
I think the static analyser is incorrect here, as it doesn't understand the function binding. Let's add asserts to make let it believe we check for the case the function has NULL value. Should not affect the binary output in the real world as asserts are only added when compiled in debug mode.

/cc @mdiluz


afayaz-feral:
Thank you for this!

Issue #287, opened 2020-09-28
`ninja install` installs a static library (libgamemodeauto.a) - how to resolve?


kakra:
I don't think that building/installation of `/usr/lib/libgamemodeauto.a` is needed, or is it for game developers to statically link gamemode into their binary? What is the proposed solution?

Gentoo QA complains about this.

See-also: https://bugs.gentoo.org/745096

stephanlachnit:
Package inih. Or set the `default_libraries` Meson variable for inih to shared.

kakra:
That's not about inih, inih is packaged separately already now in Gentoo, linked dynamically.

stephanlachnit:
Ahhh ups sry misread. Yeah it's for game devs, just like the header and pkgconfig files.

kakra:
Is there a way to make installing that .a file opt-in or opt-out? Because Gentoo QA requires that static libs are only installed when explicitly opting in through a packaging flag (useflag). Of course, I could just `rm` the file before packaging it but if it would not be built in the first place unless requested, that's probably a cleaner solution.

Also, Gentoo calls this flag `static-libs` which doesn't quite serve the purpose of this libs, IMHO, as it is meant for development. So I'd rather have it as an optional thing and make add a `devel` useflag and maybe also remove pkgconfig installation maybe? Because otherwise, the devel toolchain of gamemode would be incomplete. So, either go with everything or none, I guess.

stephanlachnit:
Hm no that isn't really possible in an easy way (one could implement a rather ugly if else statement, but IMHO the static lib certainly belongs to the development files). However we could add an option to not install any development files (but this is rather unusual, in Debian, Fefora etc all dev files are installed by default and put in a separate package).

kakra:
I'll discuss with the Gentoo devs first what the preferred approach would be, based on that I either remove just that file from packaging (unless opted in), or I'll suggest to add the option to not install devel files at all.

stephanlachnit:
> However, discarding static libs from install just for being static libs seems a bit like a sledge hammer method to me.

Usually I agree here, especially for system libs. However, most game development can't rely on system libs for compatibility reasons.

PR #378, opened 2022-08-06
Don't force installation of static libgamemodeauto

chewi:
Defining the library with `library` rather than `both_libraries` allows the user to choose which type they want to install using `-Ddefault_library`.

Closes: https://github.com/FeralInteractive/gamemode/issues/287

### DIFF
--- a/data/gamemodelist.1.in
+++ b/data/gamemodelist.1.in
@@ -1,4 +1,4 @@
-.\" Manpage for gamemoderun.
+.\" Manpage for gamemodelist.
 .\" Contact linux-contact@feralinteractive.com to correct errors or typos.
 .TH gamemodelist 1 "4 May 2020" "@GAMEMODE_VERSION@" "gamemodelist man page"
 .SH NAME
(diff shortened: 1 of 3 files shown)


stephanlachnit:
Can you please test that when running `default_library= both` and `default_library=static` libgamemode is always only a shared library and only libgamemodeauto is affected? Even then, I think it is sensible to always build libgamemodeauto static as part of the development files. Statically linking it makes perfect sense for distributed binaries of games (and it is really really tiny, it's basically only a dynamic loader for the shared library).

chewi:
I've tested and it is all as you say.

How would you like `default_library=static` to be handled? Don't install gamemoderun? Emit a warning about that?

stephanlachnit:
> How would you like default_library=static to be handled? Don't install gamemoderun? Emit a warning about that?

Maybe a warning and not installing gamemoderun / building libgamemode at all?

chewi:
Done that now. Sorry for the wait.

stephanlachnit:
Thanks, LGTM if the CI passes (cc @afayaz-feral)

Issue #377, opened 2022-08-05
Bootstrap Error: 


WickedWizard3588:
**Describe the bug**
Script `bootstrap.sh` fails with following error: 
```
╭─wickedwizard3588@HP in ~/Documents/Games took 95ms
✦  ╰─λ git clone https://github.com/FeralInteractive/gamemode.git && cd gamemode && git checkout 1.7 && yay -S meson systemd git dbus libinih && ./bootstrap.sh
Cloning into 'gamemode'...
remote: Enumerating objects: 2964, done.
remote: Counting objects: 100% (248/248), done.
remote: Compressing objects: 100% (151/151), done.
remote: Total 2964 (delta 124), reused 173 (delta 84), pack-reused 2716
Receiving objects: 100% (2964/2964), 684.65 KiB | 2.04 MiB/s, done.
Resolving deltas: 100% (2080/2080), done.
Note: switching to '1.7'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by switching back to a branch.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -c with the switch command. Example:

  git switch -c <new-branch-name>

Or undo this operation with:

  git switch -

Turn off this advice by setting config variable advice.detachedHead to false

HEAD is now at 4dc99df Specify correct inih directory after update to r54
[sudo] password for wickedwizard3588: 
warning: meson-0.63.0-1 is up to date -- reinstalling
warning: systemd-251.3-1 is up to date -- reinstalling
warning: git-2.37.1-1 is up to date -- reinstalling
warning: dbus-1.14.0-1 is up to date -- reinstalling
warning: libinih-55-2 is up to date -- reinstalling
resolving dependencies...
looking for conflicting packages...

Package (5)   Old Version  New Version  Net Change

core/dbus     1.14.0-1     1.14.0-1       0.00 MiB
extra/git     2.37.1-1     2.37.1-1       0.00 MiB
core/libinih  55-2         55-2           0.00 MiB
extra/meson   0.63.0-1     0.63.0-1       0.00 MiB
core/systemd  251.3-1      251.3-1        0.00 MiB

Total Installed Size:  70.77 MiB
Net Upgrade Size:       0.00 MiB

:: Proceed with installation? [Y/n] 
(5/5) checking keys in keyring                                                                          [#############################################################] 100%
(5/5) checking package integrity                                                                        [#############################################################] 100%
[...]

PR #379, opened 2022-08-10
Fix building when pidfd_open is available

nyz93:
On glibc 2.36 pidfd_open was made available, but doesn't work without an include.

Should fix: #377


elektronicznypank (approved):
Builds now, thanks.

nyz93:
What's blocking this from getting merged?

PR #380, opened 2022-08-11
Fix build with glibc 2.36

benjamb:
glibc 2.36 has added the pidfd_open function, if sys/pidfd.h exists then
we should include it.

Fixes the following error:

    ../common/common-pidfds.c: In function ‘open_pidfds’:
    ../common/common-pidfds.c:70:26: error: implicit declaration of function ‘pidfd_open’; did you mean ‘fdopen’? [-Werror=implicit-function-declaration]
       70 |                 int fd = pidfd_open(pid, 0);
          |                          ^~~~~~~~~~
          |                          fdopen

### DIFF
--- a/common/common-pidfds.c
+++ b/common/common-pidfds.c
@@ -47,7 +47,11 @@ POSSIBILITY OF SUCH DAMAGE.
 #include <sys/types.h>
 #include <unistd.h>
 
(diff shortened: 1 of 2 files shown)


benjamb:
Oops, looks like I missed https://github.com/FeralInteractive/gamemode/pull/379.

PR #385, opened 2022-09-13
Add Snap support for portals

ashuntu:
This change allows GameMode to call the portal dbus as opposed to daemon dbus when inside a Snap environment, the same as with Flatpak. This way, GameMode can be properly included in Snaps and used with a portal interface.

See also: https://github.com/flatpak/xdg-desktop-portal/pull/882


sergio-costas:
LGTM.

PR #393, opened 2022-10-14
Clarify that `gpu_device` must be changed in `gamemode.ini`

Calinou:
Alternative to https://github.com/FeralInteractive/gamemode/pull/391, in case we decide not to go forward with that solution.


afayaz-feral:
Closing: Out of the two solutions I prefer #391, best to use a sensible default in the first place, rather than expecting the user to understand and fix it themselves.

PR #466, opened 2024-03-07
Fix idle inhibitor closing bus connection too early

notpeelz:
According to the [Idle Inhibition Service specification](https://specifications.freedesktop.org/idle-inhibit-spec/latest/ch03.html), inhibition stops when the client disconnects from the bus:
> Inhibition will stop when the `UnInhibit` method is called, or the application disconnects from the D-Bus session bus (which usually happens upon exit).

### DIFF
--- a/daemon/gamemode-context.c
+++ b/daemon/gamemode-context.c
@@ -87,6 +87,8 @@ struct GameModeContext {
 
 	struct GameModeCPUInfo *cpu; /**<Stored CPU info for the current CPU */
 
+	GameModeIdleInhibitor *idle_inhibitor;
+
 	bool igpu_optimization_enabled;
 	uint32_t last_cpu_energy_uj;
 	uint32_t last_igpu_energy_uj;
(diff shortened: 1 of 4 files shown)


daemon/gamemode-context.c
```diff
@@ -87,6 +87,8 @@ struct GameModeContext {
 
 	struct GameModeCPUInfo *cpu; /**<Stored CPU info for the current CPU */
 
+	GameModeIdleInhibitor* idle_inhibitor;
```
afayaz-feral, line 90: This fails the clang format check, because the `*` should be beside the variable name.

daemon/gamemode-dbus.c
```diff
@@ -708,81 +708,102 @@ void game_mode_context_loop(GameModeContext *context)
...
+		return NULL;
+	}
+
+	inhibitor->bus = bus_local;
+	inhibitor->cookie = cookie;
+
+	return inhibitor;
+}
+
+void game_mode_destroy_idle_inhibitor(GameModeIdleInhibitor *inhibitor) {
```
afayaz-feral, line 778: This line also fails the clang format check, the `{` should be on a new line.

afayaz-feral (changes requested):

afayaz-feral (approved):

notpeelz:
Thanks!

PR #493, opened 2024-08-31
fix project version in meson.build (1.8.1->1.8.2)

gschintgen:
Hi, here's a minor one-line fix. When building from source I was puzzled why `bootstrap.sh` was showing 1.8.1 when I intended to install 1.8.2.


afayaz-feral:
I've updated release 1.8.2 to include this fix, and the tag has also been updated.

PR #507, opened 2024-11-07
Increase NVIDIA overclock limit

Anuskuss:
Personally I'm against this hand-holding (you already require root permissions to overclock at all) and think that the whole block of code gotta go but if you wanna keep this at least use some more realistic numbers. I can personally do 140/2790 (150 clock crashes, 2800 vram has artifacts) but I've seen people comfortably at 250 clock so 300 should be a good upper bound.
I would also suggest that instead of failing at all to just do a `max(config_get_nv_core_clock_mhz_offset(), 300)` but that would require a few more changes.

### DIFF
--- a/daemon/gamemode-gpu.c
+++ b/daemon/gamemode-gpu.c
@@ -100,8 +100,8 @@ int game_mode_initialise_gpu(GameModeConfig *config, GameModeGPUInfo **info)
 		/* Reject values over some guessed values
 		 * If a user wants to go into very unsafe levels they can recompile
 		 */
-		const int nv_core_hard_limit = 200;
-		const int nv_mem_hard_limit = 2000;
+		const int nv_core_hard_limit = 300;
+		const int nv_mem_hard_limit = 3000;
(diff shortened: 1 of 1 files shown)


afayaz-feral:
I agree that the hardcoded limit is unnecessary, and I assume the original author picked those specific values based on the hardware available at the time. Increasing the limit doesn't really solve the underlying problem.

I've been investigating the consequences of removing the limit altogether.

luisalvarado:
Depending on the nvidia card and then on top of that the model, the amount that can be overclocked for the processor and for the memory will change. For example a 4090 (Zotac AMP) can be overclocked to 220 / 400 or 200 / 500

But a 5090 Asus Astral OC which is technically less likely to be safe to be overclock can be overclocked to 350 / 550

Then a 4060 ti (Zotac) can be overclocked to 150 / 250

So the values change depending not only on the model, but the nvidia card. I do not know what the fixed values are but whatever it is, it has to be very small to be safe.

PR #510, opened 2024-11-22
Update README.md

Kaydax:
PolyMC seems to have been removed for the reason "PolyMC died" yet the project is still active. Requesting this be added back as it still supports gamemode even on the latest version


afayaz-feral:
Merged. For what it's worth, at the time of the previous change, the majority of developers had left PolyMC to form Prism. Interesting to see it's remained active!

PR #517, opened 2025-01-29
Support setting the platform profile

### LINKED ISSUE #516: Add support for platform profile
Depending on the platform profile, some other changes are unavailable (e.g., governor), so it would be good if there was a way to automatically set it (and have it correctly restored later).

https://www.kernel.org/doc/html/latest/userspace-api/sysfs-platform_profile.html

MithicSpirit:
See commit messages for more details.

This adds support for setting the platform profile via `/sys/firmware/acpi/platform_profile`. It behaves very similarly to the governor, but with just one file.

Note that setting the platform profile may change other options (like the governor) and prevent them from being altered further. Thus, this also necessitates a small refactor where the initial values are stored before any changes.

The tests for platform profile treat it as required (set the status to `-1` if they fail). However, I'm not sure whether this should be the case, as I don't know how common this feature is yet.

My nitpicking also got the best of me, so I extracted split lock mitigation stuff into `common-splitlock.c`. For now, this really only matters for the path, but, in theory, `procsysctl` could be altered to also allow getting the split lock mitigation state, rather than just setting it. Please let me know if this change is undesired, as I can just remove the last commit if so.

Closes #516

### DIFF
--- a/common/common-profile.c
+++ b/common/common-profile.c
@@ -1,0 +1,78 @@
+/*
+
+Copyright (c) 2025, MithicSpirit
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+ * Redistributions of source code must retain the above copyright notice,
+   this list of conditions and the following disclaimer.
+ * Redistributions in binary form must reproduce the above copyright
+   notice, this list of conditions and the following disclaimer in the
+   documentation and/or other materials provided with the distribution.
+ * Neither the name of Feral Interactive nor the names of its contributors
+   may be used to endorse or promote products derived from this software
+   without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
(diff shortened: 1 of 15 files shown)


afayaz-feral:
I was in the process of looking through this, and just saw that you made some further changes.
Do you want to make this PR a draft for now?

MithicSpirit:
> I was in the process of looking through this, and just saw that you made some further changes. Do you want to make this PR a draft for now?

I just forgot about `gamemoded -t` and had to add checks for that. I've been running this on my system for a few days with no problems, so unless there's another oversight like that, I don't think it needs to be a draft. Sorry about that.

common/common-profile.c
```diff
@@ -0,0 +1,78 @@
+/*
+
+Copyright (c) 2025, MithicSpirit
```
afayaz-feral, line 3: We plan on changing the copyright notice to state "the GameMode contributors" to recognise the contributions from the community.

It will get a little messy if each source file has a specific individual contributors name. In particular, this won't match the notice in the LICENSE.txt file, and reproductions of the copyright notice on binary distrubutions would be similarly complicated.

This won't affect your existing copyright on this code. Additionally, the license of the project and other code will remain the same with full attribution still available on the project page.

With all that said, could you change this line to "Copyright (c) 2025, the GameMode contributors" before I merge this pull request?

common/common-profile.h
```diff
@@ -0,0 +1,44 @@
+/*
+
+Copyright (c) 2025, MithicSpirit
```
afayaz-feral, line 3: As above, please change this line to "Copyright (c) 2025, the GameMode contributors".

common/common-splitlock.c
```diff
@@ -0,0 +1,64 @@
+/*
+
+Copyright (c) 2025, MithicSpirit
```
afayaz-feral, line 3: As above, please change this line to "Copyright (c) 2025, the GameMode contributors".

common/common-splitlock.h
```diff
@@ -0,0 +1,44 @@
+/*
+
+Copyright (c) 2025, MithicSpirit
```
afayaz-feral, line 3: As above, please change this line to "Copyright (c) 2025, the GameMode contributors".

util/platprofctl.c
```diff
@@ -0,0 +1,84 @@
+/*
+
+Copyright (c) 2025, MithicSpirit
```
afayaz-feral, line 3: As above, please change this line to "Copyright (c) 2025, the GameMode contributors".

afayaz-feral (changes requested):

PR #521, opened 2025-02-15
add RestartGamemode command

MithicSpirit:
This command allows GameMode to leave the context and then enter it again, which
re-applies all system optimizations. It is useful in cases where another
program (like TLP) may override GameMode's optimizations.

This is exposed to users by the `-R` or `--restart` flags to `gamemoded`.


MithicSpirit:
This PR is pretty much ready. I just need to do some testing and look over it one more time, so I don't blunder the same way I did for #517. Testing and feedback is appreciated.

PR #565, opened 2026-06-13
Remove unreachable code and duplicate the noreturn label

mdiluz:
Spotted when having a look at the main function - `game_mode_context_loop` is noreturn and is attributed as such at it's declaration. This duplicates that attribute over to the definition and removes redundant code.

Turns out GameMode could never "quit naturally", at least not since the dbus loop was implemented.

### DIFF
--- a/daemon/gamemoded.c
+++ b/daemon/gamemoded.c
@@ -353,9 +353,5 @@ int main(int argc, char *argv[])
 	/* Run the main dbus message loop */
 	game_mode_context_loop(context);
 
-	game_mode_context_destroy(context);
-
-	/* Log we're finished */
-	LOG_MSG("Quitting naturally...\n");
-	sd_notify(0, "STATUS=GameMode is quitting naturally...\n");
+	/* The contex loop is noreturn, exit is handled elsewhere */
 }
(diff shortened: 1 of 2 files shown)


daemon/gamemoded.c
```diff
@@ -353,9 +353,5 @@ int main(int argc, char *argv[])
 	/* Run the main dbus message loop */
 	game_mode_context_loop(context);
 
-	game_mode_context_destroy(context);
-
-	/* Log we're finished */
-	LOG_MSG("Quitting naturally...\n");
-	sd_notify(0, "STATUS=GameMode is quitting naturally...\n");
+	/* The contex loop is noreturn, exit is handled elsewhere */
```
afayaz-feral, line 356: Nitpick: typo in comment

afayaz-feral:
Looks good to me, other than that typo.
I'd like our `build_and_test` action to be working before I merge, though.

Currently trying to work out why the Ubuntu repo 404s for `libsystemd-dev`.

daemon/gamemoded.c
afayaz-feral: Dropped for now, I'll merge everything of yours so far and then see the state of the static analyser check.

